PR #66, opened 2014-12-05
Add a few more keywords and builtins

DonRyuDragoni:
Added keywords ("->", ":", "struct", "new") and builtins ("build-list", "in-range", "make-object", "remove", "remove\*", "sort", "string-append\*", "string-join", "take", "tar", "tar->output", "tar-gzip").

Should "else" be added too? Or the types for typed/racket? (Real, String, Integer, etc.)


greghendershott:
Merged -- thanks!

PR #75, opened 2015-01-07
Include the match-lambda* family and also enable newline

3rdLaw:
Added the match-lambda* family into your 'cond' clause, and also enabled newline (in addition to space) as a regexp end.


greghendershott:
Merged; thanks!

PR #82, opened 2015-01-26
smart-open-paren & builtins update

3rdLaw:
I added the *match-let* family and *named let*s to smart-open-paren. I made a separate entry for named let since it's special, and the regexp is a bit ugly since I had to allow for any proper Racket identifier as the name.

I also added a bunch of regular list functions into the builtin list from racket/list. I saw you had *take* on the list already (*take* is not in base), so I concluded the builtin list wasn't restricted to racket/base or something like that: hopefully that was a correct conclusion.


greghendershott:
Thanks! I'm sorry to be a little slow to get to this. Two big snowstorms in Boston the last week. Also I was doing some other significant updates to racket-mode.

I left a few comments about indentation and avoiding tabs. Sorry if that's nit-picky.

- Would it be possible for you to make these changes? If you do it, then the commits will show you as the author.

- Otherwise, if you're too busy, I can easily change it on my end. But after I squash it down to be one clean commit before I rebase to master and merge to master... I think the resulting commit won't show you as the author. (I'd be happy to note you in the commit message and in the contributors list, at least.)

Either way is OK for me, just please let me know which you prefer?  And thanks again for the submission, I really appreciate it!

3rdLaw:
I'll happily redo them tomorrow.

PR #86, opened 2015-02-04
 smart-open-paren & builtins updates

3rdLaw:
This is mostly the same as pull request #82 (I added a handful more builtins), but, more importantly, this time I didn't mess up the formatting:)


greghendershott:
Thanks! That diff is definitely easier to understand! :)

That's interesting I had missed so many built-ins. Originally I'd generated the list programmatically by using `namespace-mapped-symbols` on `#lang racket`. However back then I had cargo-culted how to do that and only half understood what I was doing. I'm happy you fixed it up!

3rdLaw:
I've been adding everything in the enclosing module once I notice a single builtin that's missing fontlock, but at this point we've probably included most of the non-esoteric ones:)

PR #90, opened 2015-02-10
add racket-pcomplete

kaihaosw:
When you use eshell-mode or shell-mode as your main shell, racket and raco command completion is convenient. Emacs use pcomplete for command completion, so I add this.

For now, I just add pcomplete/raco.  If you think the code should be merged, I will add pcomplete/racket.


greghendershott:
Thank you very much for this pull request. I like the idea of providing shell completion.

Before I merge, may I ask about another idea, how to do this?

It would be elegant to discover all the raco commands and their options, programmatically. That way, as users install more tools, and/or the tools add more options, it would be correct automatically (compared to updating the static list).

How to find the tools? It looks like the following returns a list of information about all installed tools:

```
(require raco/all-tools)
(all-tools)
```

How about the command-line options for each tool?  This is trickier. Two ideas:

1. Probably most tools use the [`command-line`](http://docs.racket-lang.org/reference/Command-Line_Parsing.html) form. It might not be too difficult to find that in the tool's source?  (Weak point: Not all tools use `command-line` at all. Some use it only via some fancy macros, which I think is the case for `raco pkg`.)

2. Run the tools with the `-h` flag and parse the text output.  Weak point: Parsing text is a bit kludgy. Good point: It works for all tools that provide help. If they use `command-line`, the help format is very consistent.

---

However. Maybe I'm over-thinking this. If you think this idea is too complicated or too difficult, then I should simply merge the PR as-is. :)

kaihaosw:
Hi.
  Thanks your advices.
  I came to racket recently, and I can't get the big picture of racket now. For me now, it's not convenient to type in eshell when there is no completion.
  I will make the code better when I get family with it.

PR #95, opened 2015-02-23
Added match-let* to smart-open-paren and updated builtins list

3rdLaw:
Added match-let* to smart-open-paren and a slew of builtins. 

The builtins this time are mostly from set, but also some here-and-there from sequence, regexp, string, and also a bunch of hash-based ones (I noticed a good chunk of the hash functions currently getting fontlock treatment are from MzScheme: any which start with hash-table-... instead of hash-...).


greghendershott:
Merged -- thanks!!

PR #96, opened 2015-02-24
Update README.md. Packages for racket on OSX via homebrew

manute:
When the installation of racket in OS X is made via homebrew  the repl on emacs doesn´t work. This happens because it is a minimal installation and needs more additional packages.


greghendershott:
Good point; thank you! Merged.

PR #98, opened 2015-02-24
Added Tools submenu w/ Profile, Check Syntax, and Error Trace (so far).

zenspider:


racket-mode.el
```diff
@@ -108,6 +108,10 @@ http://www.gnu.org/licenses/ for details.")
      ["Last S-Expression" racket-expand-last-sexp]
      "---"
      ["Again" racket-expand-again])
+    ("Tools"
+     ["Profile" racket-profile]
+     ["Check Syntax" racket-check-syntax-mode]
+     ["Error Trace" racket-errortrace])
```
greghendershott: Thanks for doing this. One hitch: The last/third item should be deleted. There's no `racket-errortrace` function.

There is a racket-error-level defcustom variable in racket-custom.el, and the `'high` value corresponds to using errortrace for better stack traces in error messages.

If you're thinking that should be out in a menu, and not buried in customizations? That's a good idea. Unfortunately that would mean writing a few functions, each to set one of the levels, so the menu has something to connect to. And I'm not sure if that's a slippery slope, why not other such settings....

zenspider: > On Feb 24, 2015, at 19:09, Greg Hendershott <notifications@github.com> wrote:
> 
> Thanks for doing this. One hitch: The last/third item should be deleted. There's no racket-errortrace function.

oops! Fixed. I wound up writing this into the installed package because I used elisp/emacs to find racket-run.

greghendershott:
Oh I see. `racket-run` with a prefix, like <kbd>C-u F5</kbd> or <kbd>C-u C-c C-k</kbd>. You added a function, `racket-errortrace`, so it can be a menu target. Got it.

(0) Nitpicky: What do you think about renaming that from `racket-errortrace` to `racket-run-with-errortrace`, and, adding a doc string like,

```
(defun racket-run-with-errortrace ()
  "Run with `error-context-level' set to 'high.
This is just `racket-run' with a C-u prefix. Defined as a function so it can
be a menu target."
```

? 

(1) I'd like to squash these into one commit, and rebase to master, before merging to master. But IIUC, if _I_ do so it will lose you as the author. Would you like to do the squash, and force push here?

---

But if you don't have time to do 0 or 1, let me know. Regardless, thanks again for contributing this!

greghendershott:
I edited the doc string in my previous comment. Just in case you read it via email not on the web site, and don't see the edit. Updated it's:

```
(defun racket-run-with-errortrace ()
  "Run with `racket-error-context' temporarily set to 'high.
This is just `racket-run' with a C-u prefix. Defined as a function so it can
be a menu target."
```

zenspider:
@greghendershott I just rebased and added your requests. Let me know what you think. Happy to punt on this as well if you don't think it fits any longer.

greghendershott:
Rebased onto your other PR #218 and merged; thanks!

PR #118, opened 2015-03-20
Add keywords

aidalgol:
Add class- and object-related keywords to racket-keywords.

I have omitted trait forms because I am not sure whether they should be
considered keywords or builtins for the purposes of racket-mode.


greghendershott:
Merged; thanks!

PR #129, opened 2015-04-22
Add `field` keyword

aidalgol:
Add `field` to `racket-keywords`.  This should have been in PR #118, but I somehow missed it.  Sorry about that.


greghendershott:
Thanks for adding this.

So, I can't merge as-is, due to a couple problems:

1. Did you mean to _rebase_ your `syntax-highlight-fixes` branch to `master`? (While you're at it, a rebase plus _squash_ down to one commit would be nice. Really this should amount to one commit with that one line added, I think?)

2. Travis CI is reporting failing tests because making `field` a Racket keyword causes font-lock to change its appearance in either or both of the example/*.rkt files. As [the comments](https://github.com/greghendershott/racket-mode/blob/master/example/example.rkt#L3-L4) in those files explain, you can <kbd>M-x faceup-write-file</kbd> to generate new .faceup files for the test.

However. Given this is all for a one-line change, I could do this on my end. Please let me know which you prefer, for you to do it or me?  (And either way, thanks again for helping improve racket-mode in the past and also now!)

aidalgol:
I probably should have deleted the branch after my previous pull request was accepted, but deleting the branch with an open pull request was almost certainly the wrong thing to do.

I have deleted the branch and rebranched off master, and re-created this PR: #130.

PR #141, opened 2015-06-13
Fix void `#'racket-unicode-input-method-enable`

vkz:
It seems you simply forgot to require it, so things like this don't work:
```elisp
(add-hook 'racket-mode-hook      #'racket-unicode-input-method-enable)
```

### DIFF
--- a/Reference.md
+++ b/Reference.md
@@ -23,11 +23,15 @@
 
 Save and evaluate the buffer in REPL, much like DrRacket's Run.
 
+If point is within a submodule form, the REPL "enters" that
+submodule (uses its language info and namespace).
+
 When you run again, the file is evaluated from scratch -- the
 custodian releases resources like threads and the evaluation
 environment is reset to the contents of the file. In other words,
 like DrRacket, this provides the predictability of a "static"
-baseline, plus some interactive exploration.
+baseline, plus the ability to explore interatively using the
+REPL.
 
 See also [`racket-run-and-switch-to-repl`](#racket-run-and-switch-to-repl), which is even more like
 DrRacket's Run because it selects the REPL window (gives it the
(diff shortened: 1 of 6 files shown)


greghendershott:
Thanks!

Dang. I thought this was already handled by the [`;;;###auto-load`](https://github.com/greghendershott/racket-mode/blob/master/racket-unicode-input-method.el#L23).

I should probably just merge your PR, but first I want to re-read that chapter of the Elisp docs and make sure I understand.

greghendershott:
Sorry to leave this hanging. I got part way into investigating this, then suddenly got yanked into dealing with some other things a few weeks ago.

So I need to find the plot again, and make sure, but:

I _think_ what I learned is that actually it works fine when racket-mode is installed as a package from MELPA (I tried on a fresh machine). I _think_ you're seeing a problem only because you're using racket-mode from source, and therefore `update-file-autoloads` isn't run.

If so, then adding a `require` might be "too heavy", and instead we should ask source users to `M-x update-directory-autoloads`?  I think??

greghendershott:
> instead we should ask source users to M-x update-directory-autoloads? I think??

In which case this PR would, I guess, change into one updating README?

vkz:
sure, fine with me. I'm yet to come to grips with autoloads machinery. One question is weather this implies the need to `update-directory-autoloads` every time we restart emacs if we run racket-mode from source, or add this line to the racket-mode section in our init.el? Readme wouldn't go amiss that's for sure. Thank you

greghendershott:
I did some more reading. I think the answer is simply that you should do a `make compile` every time you update from git.

That creates `.elc` files; when Emacs evals them, autoloads are executed.

Can you confirm that works for you, too?  If so I'll update the README.

---

p.s. This took me awhile to realize because I only `make compile` to ensure I get a clean compile, then almost always immediately `make clean`. But I do this because I'm pretty much constantly updating racket-mode.

Issue #41, opened 2014-09-10
Offer to help with bytecode mismatches?


greghendershott:
When working on a project that's been compiled -- that has `compiled` subdir(s) created by `raco make` or `raco setup` -- eventually you can get this error:

```
link: module mismatch; possibly, bytecode file needs re-compile because dependencies changed
```

The solution is to `rm -rf compiled/` or re-run `raco make` or `raco setup`. Which isn't _difficult_ to do. But it's kind of a buzz-kill flow-breaker to need to `M-x shell`, `cd` to the right directory, issue the `rm` or `raco` command... then remember what you were focused on before this annoying distraction.

Could `racket-mode` notice this sort of error, and do something automatically?  Maybe after a y/n prompt?

greghendershott:
See https://github.com/rmculpepper/custom-load

cpitclaudel:
Instead of going to a separate shell, I would expect users to use M-x compile and put raco make in there. We could help by setting `compile-command` to `raco make` by default:

```
compile-command is a variable defined in ‘compile.el’.
Its value is "make -k "

  This variable may be risky if used as a file-local variable.
  This variable is safe as a file local variable if its value
  satisfies the predicate ‘(lambda (a) (and (stringp a) (or (not (boundp (quote compilation-read-command))) compilation-read-command)))’.

Documentation:
Last shell command used to do a compilation; default for next compilation.

Sometimes it is useful for files to supply local values for this variable.
You might also use mode hooks to specify it in certain modes, like this:

    (add-hook 'c-mode-hook
       (lambda ()
	 (unless (or (file-exists-p "makefile")
		     (file-exists-p "Makefile"))
	   (set (make-local-variable 'compile-command)
		(concat "make -k "
			(if buffer-file-name
			  (shell-quote-argument
			    (file-name-sans-extension buffer-file-name))))))))

You can customize this variable.
```

PR #144, opened 2015-07-04
Bind C-c C-c to #'racket-run

cpitclaudel:
This is in line with the practice of other major modes, like
python-mode. Some other modes (like haskell-mode) use this to map to a
mode-specific version of #'compile, but I'm not sure that this would be
too relevant here. What do you think?

### DIFF
--- a/racket-mode.el
+++ b/racket-mode.el
@@ -56,6 +56,7 @@ http://www.gnu.org/licenses/ for details.")
 (defvar racket-mode-map
   (racket--easy-keymap-define
    '(("C-c C-k"     racket-run)
+     ("C-c C-c"     racket-run)
      ("C-c C-z"     racket-repl)
      ("<f5>"        racket-run-and-switch-to-repl)
      ("M-C-<f5>"    racket-racket)


greghendershott:
Good point. I've wondered about this.

Racket does have `raco make` to create bytecode `zo` and `dep` files under `compiled/`.  But in my experience usually this gets run as part of `raco setup` or `raco pkg install` -- not directly as a `raco make` of a specific file.

So if we were to bind <kbd>C-c C-c</kbd> to something "compilation related", where it saves people needing to go to a shell,  maybe it should be `raco setup`, re issue #41 ? But I'm not sure....

cpitclaudel:
I've added a comment there :) I don't think C-c C-c should be made redundant with a properly configured M-x compile. I also expect racket-run to be much more common that compiling.

I use C-c c to run compile, and that has been pretty convenient. That could be suggested in the docs somewhere.

greghendershott:
1. Sorry: I confused things by mentioning issue #41 in connection with this. When issue #41 arises, it's almost always because some _other_ file(s) need to be recompiled. Closing issue #41 would entail doing something like what I mentioned in the first comment -- using https://github.com/rmculpepper/custom-load.

2. I do like your comment -- which I misled you to leave over there :) -- about setting Emacs `compile-command` to `raco make` in racket-mode buffers. I'll do that.

3. Once I do 2, is there anything left to-do on this issue?  Should we leave it up to individual users to bind <kbd>C-c C-c</kbd> if/as they prefer?  Or should we bind it to `compile` in racket-mode as a reasonable default?  What do you think?

cpitclaudel:
Cool for 2. For 3, I would actually vote for biding C-c C-c to the same as C-c C-k, as it matches other language modes like Python-mode. I wouldn't bind it to compile, because users often already have a binding for that, and because it would be slightly inconsistent with other language modes. Of course, that wouldn't prevent users from rebinding C-c C-c.

Anyhow, this is a minor thing; it just bit me when I first tried the mode :)

greghendershott:
OK. Done. Thanks for the PR!

PR #145, opened 2015-07-04
Fix condition around save-buffer in racket--do-run

cpitclaudel:
It's not enough to check if a file has been modified; the underlying
file could have been deleted, or the buffer could have never been
saved (it could have been generated by a command, or it could be an
empty buffer). The added check is the same as the one Emacs uses when
calling `save-buffer`. The fix may or may not need to be extended to
`racket-trim-requires` and `racket-save-requires` (I would guess not
really).


greghendershott:
Good point. Merged. Thanks!

PR #146, opened 2015-07-05
Indent rule for rackunit's test-case, test-begin

dpercy:
This adds an indentation rule for rackunit's `test-case` and `test-begin` forms:
http://docs.racket-lang.org/rackunit/api.html?q=rackunit#%28part._.Test_.Cases%29

If you like the change, do you want an example added to test it?


greghendershott:
Merged. Thanks!

PR #177, opened 2015-11-11
racket--version doesn't quote argument properly on Windows

freakhill:
racket-mode

For windows support.

### DIFF
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -129,7 +129,7 @@ Never changes selected window."
 (defun racket--version ()
   "Get the `racket-racket-program' version as a string."
   (with-temp-buffer
-    (shell-command (concat racket-racket-program " -e '(version)'")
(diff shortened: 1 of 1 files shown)


greghendershott:
Thank very much for reporting this problem, as well as submitting a PR with a proposed fix.

After re-reading the Emacs documentation, I think the more basic problem is that I used `shell-command`. Its doc string says:

> In Elisp, you will often be better served by calling `call-process` or
`start-process` directly, since it offers more control and does not impose
the use of a shell (with its need to quote arguments).

For example instead using `call-process`:

```cl
(call-process racket-racket-program
                  nil   ;infile
                  t     ;destination: current-buffer
                  nil   ;redisplay
                  "-e"
                  "(version)")
```

Although I appreciate you taking the time to submit a PR for the issue -- thank you! -- I think instead I'll use `call-process`.

freakhill:
thanks!

PR #190, opened 2015-12-18
Make magic [ paren key work in Racket REPL buffers

david-christiansen:
Now, the Paredit bindings that check whether a buffer is in `racket-mode` also accept buffers in `racket-repl-mode`.


greghendershott:
Merged; thanks!

PR #192, opened 2015-12-30
Add "lambda" and "define" to smart [ key forms

### LINKED ISSUE #191: The smart [ key inserts parens rather than [ in argument lists
My understanding is that Racket style encourages nested parens in argument lists to be square brackets. For instance:
```
(lambda (x y [z 15] #:key [which-thing #f]) ...)
```
This can be arranged by adding `lambda` and `define` to the `let`-like forms in `racket-common.el`, but I'm a bit concerned that I may be breaking something else there.

david-christiansen:
This causes nested parens in argument lists to get [] instead of ().

Fixes #191.

I'm not completely sure that this is the Right Way, as there's lots of subtlety in the argument list syntax, but I figured I'd submit my test code rather than just describing it!

### DIFF
--- a/racket-common.el
+++ b/racket-common.el
@@ -326,7 +326,8 @@ With a prefix, insert the typed character as-is."
       ;; In addition to the obvious suspects with 'let' in the name,
       ;; handles forms like 'parameterize', 'with-handlers', 'for',
       ;; and 'for/fold' accumulator bindings.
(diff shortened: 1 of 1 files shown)


greghendershott:
I apologize it took me so long to get to this!

This looks good to me. Thank you very much for contributing it. Merged!

PR #193, opened 2016-01-03
Add smart-[-key info for object forms

david-christiansen:
Now, a style similar to that of the documentation results when using the
smart [ key to write Racket code with classes.


greghendershott:
Merged -- thanks!

david-christiansen:
You're welcome! Thanks for `racket-mode`, which I really enjoy a lot!

PR #216, opened 2016-06-29
Add function to insert ∃ for contracts

zv:


greghendershott:
Thank you for submitting this and I'm sorry to be slow to reply.

This _seems_ redundant with the existing Unicode input option described [here in the REAMDE](https://github.com/greghendershott/racket-mode#unicode-input-method) and [here in the online reference](https://github.com/greghendershott/racket-mode/blob/master/Reference.md#racket-unicode-input-method-enable).

Specifically `∃` is implemented [here in racket-unicode-input-method.el]( https://github.com/greghendershott/racket-mode/blob/master/racket-unicode-input-method.el#L194).

Would that work for you?

greghendershott:
By the way, I notice now that both the README and the Reference / doc string neglect to mention that you need `(require 'racket-unicode-input-method)` to make `racket-unicode-input-method-enable`. Both doc locations should be updated.

david-christiansen:
Another possibility is to use https://github.com/david-christiansen/dr-racket-like-unicode for a real DrRacket-like experience &lt;/shameless-self-promotion&gt;

Issue #221, opened 2016-08-15
racket-mode provides 1 instead of 0 command-line arguments


toothbrush:
Hello.  When trying to write a module to serve as a command-line entry-point to my program, i found that racket-mode provides 1 command-line argument when running a module, instead of 0, like DrRacket does.  MWE:

```racket
#lang racket
(command-line #:args () '())
```
Using racket-mode i receive the following error (the typo is actually exactly as i see it):
```racket
; run.rkt: expects no arguments on the command line, given 1 argument: /tmp/racket-mode-command-ouput-file-310598cN 
```

Whereas when loading the file (evaluating using <kbd>F5</kbd> as usual) in DrRacket it works, simply outputting a `'()`.  Someone on IRC last week (would love to provide attribution but my lack of AJAX-laden website-fu is preventing me from making botbot.me do my bidding) pointed me to this snippet: https://github.com/greghendershott/racket-mode/blob/master/run.rkt#L28 -- perhaps `(current-command-line-arguments)` should be set to `'#()` thereabouts?  Forgive me if i am saying something unforgivably dumb.  Awesome mode, thanks.

EDIT: i am using the current latest MELPA version of racket mode, `20160804.1650`.

greghendershott:
I'm sorry to be slow to reply, I've just been swamped. I'll try to look at this within the next few days. I agree it should be improved to remove the command-line argument that racket-mode itself supplies.

Also maybe racket-mode should have a feature to let you supply example args, much like DrRacket does.

PR #223, opened 2016-08-28
Provide no command-line arguments in REPL

toothbrush:
* run.rkt: Empty the vector of command-line arguments to match behaviour
  of DrRacket when evaluating modules which reference
  `current-command-line-arguments`.  Addresses #221.

### DIFF
--- a/run.rkt
+++ b/run.rkt
@@ -25,7 +25,8 @@
     (file-stream-buffer-mode (current-output-port) 'none))
   (display (banner))
   (flush-output)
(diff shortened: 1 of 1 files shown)


greghendershott:
Thank you for the pull request!

I ended up not using this because I wanted to change the `current-command-line-arguments` parameter only for the REPL's thread. Also I wanted to provide a way for the user to supply command-line arguments from Emacs, as discussed in #221.

PR #226, opened 2016-09-20
Fix error using eval when REPL in separate frame

jgarvin:
get-buffer-window will return nil if the window with the requested name
is not part of the current frame. I tend to use emacs frames instead of windows so that I can manage all of my windows emacs and non-emacs alike the same.


jgarvin:
No problem. I've pushed a new version that fixes the other locations.

greghendershott:
Thanks! I squashed it to one commit 1166035 and merged.

PR #234, opened 2016-10-27
maybe-load-language-info: load cr-submod regardless of lang-info

willghatch:
Currently a configure-runtime submodule is only loaded if module->language-info can return something.  But you can have a configure-runtime submodule even if #%module-begin doesn't have a language-info syntax property.

This patch just moves the conditional configure-runtime submodule load out of the conditional that depends on module->language-info.


greghendershott:
Thanks -- merged!

PR #239, opened 2016-11-01
fix configure-runtime submodule handling nested submodules

### LINKED ISSUE #238: Can't load test submodule on latest master
In the latest `racket-mode`, when I put the following in a buffer that's saved to a file called `new-test.rkt`:
```
#lang racket

(module+ test
  (define x 54))
```
and then `C-c C-t`, I get the following in my `*Racket REPL*` buffer:
```
; module-declared?: contract violation
;   expected: (or/c module-path? module-path-index? resolved-module-path?)
;   given: '(submod (submod #<path:new-test.rkt> test) configure-runtime)
```
Previously, it would put the REPL in the context of the test submodule.

willghatch:
Fix #238 -- sorry about that.

### DIFF
--- a/run.rkt
+++ b/run.rkt
@@ -147,7 +147,10 @@
       ((dynamic-require (vector-ref config 0)
                         (vector-ref config 1))
        (vector-ref config 2))))
-  (define cr-submod `(submod ,path configure-runtime))
+  (define cr-submod `(submod ,@(match path
+                                 [(list 'submod sub-paths ...) sub-paths]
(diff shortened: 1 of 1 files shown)


willghatch:
Ok -- I've fixed the error I made in haste.  This works for normal module paths as well as submodule paths.  Unless there is some weird require spec type that you can't put inside a submod require spec (aside from submod itself), it should work fine.

greghendershott:
Thank you! I'll merge after the Travis tests finish.

PR #242, opened 2016-12-09
Escape messages from racket-check-syntax-mode

### LINKED ISSUE #241: % signs in tooltips causes error
I'm currently working on a proof tactic language embedded in Racket's macro expander. Part of what it does is displaying tooltips that contain proof states, which contain fully-expanded syntax. So far, so good - Emacs won't display the output of the online check syntax logger, but it I can collect tooltips and attach them to the full expansion.

When using `racket-check-syntax-mode`, however, `message` is invoked to show the contents of DrRacket tooltips. Fully-expanded Racket contains lots of `%` characters, and `message` expects a format string. This does not go over well, resulting in backtraces like this one:
```
Debugger entered--Lisp error: (error "Not enough arguments for format string")
  message("nat-equality:\n1. x : (#%app ≡6 (#%app Nat1) (quote 2) (quote 3))\n0. x : (#%app Nat1)\n⊢ (#%app ≡6 (#%app U2 (quote 0)) (#%app Nat1) (#%app Nat1))\n\n")
  racket--point-entered(1 5401)
  racket--check-syntax-start()
  racket-check-syntax-mode(toggle)
  funcall-interactively(racket-check-syntax-mode toggle)
  call-interactively(racket-check-syntax-mode record nil)
  command-execute(racket-check-syntax-mode record)
  helm-M-x(nil "racket-check-syntax-mode")
  funcall-interactively(helm-M-x nil "racket-check-syntax-mode")
  call-interactively(helm-M-x nil nil)
  command-execute(helm-M-x)
```
i think that the string needs to be escaped, but I can't find a built-in Emacs function to do that.

david-christiansen:
Prevent Elisp format strings originating in Racket tooltips from
confusing message by escaping them first.

Fixes #241.

### DIFF
--- a/racket-edit.el
+++ b/racket-edit.el
@@ -595,7 +595,7 @@ location of the first one."
 
 (defun racket--point-entered (old new)
   (pcase (get-text-property new 'help-echo)
-    (s (message s)))
+    (s (message "%s" s)))
   (pcase (get-text-property new 'racket-check-syntax-def)
     ((and uses `((,beg ,end) . ,_))
      (pcase (get-text-property beg 'racket-check-syntax-use)


david-christiansen:
I'm not entirely sure that this is the right way to do this.  The docstring for `message` says to use `(message "%s" val)` to avoid treating `val` as a format string, so that seems to check out, but I'm not sure that `message` should be called here at all.

This is because Emacs already has a way of showing the `help-echo` property. First off, it's a tooltip, but it also appears in the echo area when `C-h .` is issued. Furthermore, it can be automatically displayed in the echo area when point hangs out over the area for a bit by setting `help-at-pt-display-when-idle` to `t`, according to [the manual](https://www.gnu.org/software/emacs/manual/html_node/emacs/Help-Echo.html). Doesn't this cover the use case? If so, might it not make sense to simply remove that `pcase` expression from `racket--point-entered`?

greghendershott:
Thanks for the PR!

> I'm not sure that `message` should be called here at all.

Background/motivation: I wanted the information to be displayed when navigating with the keyboard (so people don't need to grab the mouse and point). Since I was already doing `point-entered` and `point-left` handling for the color highlighting, it seemed easy to add this, too. And IIRC I couldn't find a "show tooltip" command, so I just used `message`.

Having said all that: This might be yet another example of me re-inventing a wheel that I didn't realize Emacs already invented -- maybe there's a general "show tooltips in the echo area" setting or minor mode. If so, maybe I should turn it on automatically in `racket-check-syntax-mode` so that still "just works". Or, maybe not and I should just document this for people who care.

david-christiansen:
This makes sense. `help-at-pt-display-when-idle` is /almost/ the setting you want, but not quite, because it waits for an idle timer before displaying the tooltip text, rather than doing it immediately upon point entering the region with the help text. So I suppose that the PR does make sense. 

Perhaps it should also be guarding the call the `message` so it only does it if there's actually help text present, to avoid a bunch of empty messages.

greghendershott:
I wanted to merge this today, but tried it and was surprised to notice that neither the `message` nor the overlays were displaying for me in Emacs 25.1.

From searching and reading, it seems this is because in Emacs 25 the `point-entered` and `point-left` properties have been obsoleted in favor of a new `cursor-sensor-functions` property, and a new `cursor-sensor-mode` to make it work.

Not just deprecated; in fact, the `inhibit-point-modification-hooks` variable defaults to `t`. The old `point-x` properties only work if I do a `(setq-local inhibit-point-motion-hooks nil)` in the racket-mode buffer while check-syntax-mode is active.

- Short-term I guess I should do that -- `(setq-local inhibit-point-motion-hooks nil)` -- so it still works in Emacs 25.

- Mid-term (until racket-mode can require Emacs 25+ someday), I guess I am supposed to check the Emacs version and implement both old and new?

greghendershott:
So to be clear, I'll do two things:

1. The change in this PR, enhanced with your suggestion to check for empty strings.
2. `(setq-local inhibit-point-motion-hooks nil)` to keep the old point-motion mechanism working (for now) with Emacs 25+.

PR #250, opened 2017-01-29
Add unicode for Union and Intersection.

hidaris:
After [this commit](https://github.com/racket/typed-racket/commit/aa1d36f44e0ef49d263cf53f97e578c1b58dca31), we have aliases for Union type and Intersection type.


greghendershott:
Thank you!

Issue #188, opened 2015-12-14
Inline images don't work on Windows


interplinternet:
I've tried googling and going through the source myself, but haven't found any solution. Emacs on Windows can support images fine, and I can display them in the scratchpad using what seems to be the same function used in racket-repl.el: `(insert-image (create-image [filename]))`. I've also checked my temp dir, and there is in fact an image there with the racket-generated filename. However, M-x `racket-view-last-image` gives me the message that "There aren't 1 images". Using M-: to eval `racket-image-cache-dir` gives a result of nil. I'm not sure if that's the desired output, but maybe it'll be helpful. The image is also listed in the repl as `#<Image: C:\\Users\\user\\AppData\\Local\\Temp\\racket-image-14501363761450136376317.png>`, so I thought it might be a problem relating to the escaped backslashes, but using that file's location as the target for the above insert-image+create-image function (including backslashes) still displays the image. Here's the result of `racket-bug-report`.

```
((emacs-version "24.5.1")
 (emacs-uptime "5 minutes, 25 seconds")
 (system-type windows-nt)
 (major-mode racket-repl-mode)
 (racket--source-dir "c:/Users/argos/.emacs.d/elpa/racket-mode-20151207.1952/")
 (racket-racket-program "c:\\Program Files\\Racket\\Racket.exe")
 (racket-raco-program "c:\\Program Files\\Racket\\raco.exe")
 (racket-memory-limit 2048)
 (racket-error-context medium)
 (racket-history-filter-regexp "\\`\\s *\\S ?\\S ?\\s *\\'")
 (racket-images-inline t)
 (racket-images-keep-last 100)
 (racket-images-system-viewer "display")
 (racket-pretty-print t)
 (racket-indent-curly-as-sequence t)
 (racket-indent-sequence-depth 0)
 (racket-pretty-lambda nil)
 (racket-smart-open-bracket-enable nil))
(enabled-minor-modes
 (auto-composition-mode)
 (auto-compression-mode)
 (auto-encryption-mode)
 (auto-fill-mode)
 (auto-save-mode)
 (blink-cursor-mode)
 (electric-indent-mode)
 (file-name-shadow-mode)
 (font-lock-mode)
 (global-font-lock-mode)
 (line-number-mode)
 (menu-bar-mode)
 (mouse-wheel-mode)
 (tool-bar-mode)
 (tooltip-mode)
 (transient-mark-mode))
(disabled-minor-modes
 (abbrev-mode)
[...]

PR #267, opened 2017-06-04
Modified racket-repl--replace-images regex to capture windows paths

MRFaria:
This fixes the issue with viewing inline images in the repl on windows.

I don't have a Linux machine, so I can't test that the regex string still works on a different OS.

### DIFF
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -386,7 +386,7 @@ images in 'racket-image-cache-dir'."
   (with-silent-modifications
     (save-excursion
       (goto-char (point-min))
-      (while (re-search-forward "\"#<Image: \\([-+./_0-9a-zA-Z]+\\)>\"" nil t)
+      (while (re-search-forward "\"#<Image: \\([-+./_0-9a-zA-Z:\\]+\\)>\"" nil t)
         ;; can't pass a filename to create-image because emacs might
         ;; not display it before it gets deleted (race condition)
         (let* ((file (match-string 1))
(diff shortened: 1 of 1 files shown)


greghendershott:
Thanks!

This is a good, conservative fix to the original regular expression. I might merge this.

---

On the other hand, the original regular expression -- my fault! -- smells bad to me now. Maybe instead of tweaking it, we should replace it.

I think something less fussy, like `"\"#<Image: \\([^>]+\\)>\""`, would avoid getting into OS specifics.

I also don't think it would be likely to get false matches. But if that's a worry, the regexp could match based on the fact that racket-mode's "back end" generates filenames using this pattern:

```racket
(make-temporary-file "racket-image-~a.png")
```

So, instead of merging this as-is, I _might_ give you such a regexp and ask you to confirm it also works fine for you on Windows?

MRFaria:
Yeah sure, I also think making the regexp more general is the way to go.  I can go ahead and test on windows if you have any suggestions.

greghendershott:
Great. This works for me on macOS:

```elisp
    "\"#<Image: \\(.+racket-image-.+\\.png\\)>\""
```

Does it work for you on Windows?

If yes: Would you like to update your PR with that, and I'll merge?

MRFaria:
Nice, this works for me on windows as well. I've updated the pull request

greghendershott:
Great! I'll merge once Travis CI finishes the check in a few minutes.

This will close issue #188.

greghendershott:
Thank you!

PR #268, opened 2017-06-06
racket-doc: Use net/url to encode url in find-help/mac

kalbr:
Previously, `find-help/mac` would fail to display files in `/Applications/Racket v6.9/...` i.e. the default installation path. It looks like the space needed to be escaped. The browser was Firefox. 

I fixed it before filing a ticket, hopefully this is okay.

### DIFF
--- a/help.rkt
+++ b/help.rkt
@@ -3,6 +3,7 @@
 (require (only-in help/help-utils find-help)
          (only-in help/search perform-search)
          json
+         net/url
          racket/contract
          racket/file
          racket/format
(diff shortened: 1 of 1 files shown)


greghendershott:
1. Thank you!

2. I'm embarrassed this snuck by. A long time ago I got in the habit of renaming the `Racket vN.N` folders to `Racket_vN.N`. Maybe I should stop doing that.

3. It looks like the Travis CI tests are catching the fact that `net/url-string` wasn't added until Racket 6.3 (racket-mode supports back to Racket 6.0).

    Would you like to push a commit that uses another approach (maybe `net/uri-codec`), _or_, have me do that?

    (I like the idea of the commit having your name on it, as credit for helping. But I don't mind doing it if you don't have time.)

kalbr:
Ah, sorry! My original PR used `net/url`, which Travis accepted. I switched to `net/url-string`when I realised that `net/url` did much more than encode urls, but didn't stick around long enough to see the glaring red failed check. I can either 

1. use `net/url` again. In fact, I've already pushed the old commit, or

kalbr:
Another attempt at a reply; this time with a full stomach:
 
1. My original PR required `net/url` instead of `net/url-string` and passed the CI check. The change was an optimisation anyway, so I've pushed the original commit again. No red circles are visible on my side. It looks like the functionality that I used was provided by `net/url` even before `net/url-string` was factored out.

2. Sorry for not seeing the failing Travis CI checks last time.

3. If `net/url` is too big a module to pull in, I'll have a go at using `net/uri-codec`.

4. Thanks for working on this! `racket-mode` is fantastic. My only gripe is that the `README.md` is too modest!

greghendershott:
Awesome and thanks again!

PR #279, opened 2017-09-11
Don't cdr an empty srclocs list in display-srclocs; fixes #278

### LINKED ISSUE #278: Error display handler exception with null exn:fail:read-srclocs
If an `exn:fail:read` has empty list as its srclocs, the display handler raises an exception, because it tries to `cdr` the empty list. The documentation for `exn:fail:read` doesn't claim that this list should be non-empty.

```racket
main.rkt﻿> (raise (exn:fail:read "..."
                                (current-continuation-marks)
                                '()))
; ...
exception raised by error display handler: cdr: contract violation
  expected: pair?
  given: '(); original exception raised: ...
```

Additionally, the output message here seems to be slightly malformed (semicolon missing before some lines)

iitalics:


greghendershott:
Thanks!

Your patch is a good fix to the code I wrote! But I don't love how I wrote it back then. These days I'd probably use `match*` to both (a) check the exn type and (b) safely deconstruct the list:

```racket
(define (display-srclocs exn)
  (when (exn:srclocs? exn)
    ;; Some exceptions like exn:fail:read? include the first srcloc in
    ;; exn-message, so don't show it twice. In fact, some exceptions
    ;; like exn:fail:syntax? with Typed Racket include _all_ in
    ;; exn-message in which case don't show _any_.
    (define srclocs
      (match* [exn ((exn:srclocs-accessor exn) exn)]
        [[(or (? exn:fail:read?) (? exn:fail:contract:variable?)) (cons _ xs)] xs]
        [[(? exn:fail:syntax?)                                    _          ] '()]
        [[_                                                       xs         ] xs]))
    (for ([srcloc (in-list srclocs)])
      (display-commented (source-location->string srcloc)))))
```

But, what do you think? Is this clear, or is it too dense/clever?

greghendershott:
Another way is plain `match` with `#:when` guards:

```racket
    (define srclocs
      (match ((exn:srclocs-accessor exn) exn)
        [(cons _ xs)
         #:when (or (exn:fail:read? exn) (exn:fail:contract:variable? exn))
         xs]
        [_
         #:when (exn:fail:syntax? exn)
         '()]
        [xs xs]))
```
---

Or, if you don't really case and just want me to merge the patch, feel free to say so! 😄

iitalics:
I think the second variant with `#:when` guards is the most clear. Is there a specific reason we check for both `exn:fail:read?` and `exn:fail:contract:variable?`?

greghendershott:
> Is there a specific reason we check for both `exn:fail:read?` and `exn:fail:contract:variable?`?

For the reason in the comment:

```racket
    ;; Some exceptions like exn:fail:read? include the first srcloc in
    ;; exn-message, so don't show it twice. In fact, some exceptions
    ;; like exn:fail:syntax? with Typed Racket include _all_ in
    ;; exn-message in which case don't show _any_.
```

> I think the second variant with `#:when` guards is the most clear.

OK. Would you like to update your PR, or, stick with your original commit?

(Either is fine -- I really appreciate your help and time!)

iitalics:
OK I made the changes

greghendershott:
Thank you!

PR #282, opened 2017-10-09
Make racket-run connect asynchronously.

technomancy:
Other commands which need a connection to the racket command process
will still connect synchronously, but at least it's possible to make
the initial connection (which is usually the worst source of waiting)
without blocking the entire Emacs process.

### DIFF
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -1,4 +1,4 @@
-;;; racket-repl.el
+;;; racket-repl.el -*- lexical-binding: t -*-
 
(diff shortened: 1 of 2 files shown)


greghendershott:
Thank you for submitting this!  Around the time you did, I noticed it wasn't passing tests; I was in a busy period where I couldn't dig in to figure out why. 

I ended up doing something similar in commit 1ec3c4f so I'll close this now. Thanks again!

Issue #259, opened 2017-04-08
Text displayed after read-only prompt is wrongly read-only


greghendershott:
Given a file:

```racket
#lang racket/base
(log-fatal "foo")
```

<kbd>C-c C-k</kbd> it.  The REPL shows:

```
foo.rkt﻿> ; [fatal] foo
|
```

Where `|` represents point.

Normally you should be able to type e.g. `42` and hit <kbd>RET</kbd>. But attempting to type something displays `"Text is read-only"` in the echo area.

Same with:

```rkt
#lang racket/base
(thread (λ () (displayln "hi")))
```

In other words the issue arises when Racket writes the prompt, then writes more output on the same line as the prompt.

---

This seems to be due to `comint-mode` placing a `read-only` property with the value `'fence` on the newline before point.

---

This was introduced by commit cdf6583 to close #245. I might need to revert that commit, until/unless I can't figure out why `comint-mode` does this and how to avoid it.

enjelbeteu:
`
#lang racket
(define n (read))
`
also read-only

greghendershott:
@enjelbeteu Maybe I'm misunderstanding what you're describing -- the following interaction in the REPL works fine for me:

```
foo.rkt﻿> (define n (read))
12
foo.rkt﻿> n
12
foo.rkt﻿> 
```

What are you doing, that causes a read-only message?

enjelbeteu:
@greghendershott
Given a file:
`#lang racket

(define n (read))
`
also cause read-only

greghendershott:
I went down the `comint` rabbit hole for awhile but couldn't figure out how to work around its `'readonly` `'fence` thing.

I don't think I can spend more time on this now, so, I'm going to revert commit cdf6583 and reopen #245.

PR #292, opened 2018-01-24
Make prompt characters read-only

### LINKED ISSUE #245: prevent editing the prompt in racket-repl-mode
In `racket-repl-mode`, the prompt can be edited and moved into, which leads to some occasionally curious behavior. In Python's repl mode (for example), this doesn't happen. I think, but am not sure, that the reason why has to do with the `inhibit-line-move-field-capture` text property (and maybe some other related properties): http://emacs.stackexchange.com/questions/13592/how-does-comint-mode-override-beginning-of-line-behavior-i-want-same-functional

In any case, it would be nice if the prompt were uneditable and couldn't be moved into, at least not by accident. For example, `M-b` (`backward-word`) at the prompt moves the cursor backward into the prompt.

YourFin:
Closes #245

### DIFF
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -473,7 +473,7 @@ With prefix arg, open the N-th last shown image."
   (racket--common-variables)
   (setq-local comint-prompt-regexp (rx (regexp "^[^>\n]*") "\ufeff> "))
   (setq-local comint-use-prompt-regexp t)
-  (setq-local comint-prompt-read-only nil)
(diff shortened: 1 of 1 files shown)


greghendershott:
Thank you for submitting this. However, won't this re-open #259 ?

YourFin:
Hmm sorry I hadn't dug into the derivative/parent issues as they appeared closed. Let me do some more digging tonight and I'll get back on this.

PR #295, opened 2018-03-12
Fix doc for racket-here-string-face.

pmatos:


greghendershott:
Thank you! However, `Reference.md` isn't edited directly -- it's generated from the doc strings in Emacs Lisp code.

So instead, please:

1. Make the edit [here](https://github.com/greghendershott/racket-mode/blob/master/racket-custom.el#L294).
2. Run `make doc`.
3. Now git should show that both `racket-custom.el` and `Reference.md` have changed, and you can commit.

greghendershott:
Actually I'm going ahead and just doing this myself -- thank you very much for noticing the error and providing a fix PR!

PR #299, opened 2018-04-01
Only remove overlays created by racket-mode

hkrish:
This is just a fix to the call remove-overlays. It expects a name /and/ a value to identify overlays. 
Right now racket-mode clears some overlays created by other minor modes as well.


greghendershott:
Thanks!

PR #302, opened 2018-05-24
Adds `racket-run-main` to run the `main` submodule

lenary:
`racket-run-main` will run the `main` submodule of the current
submodule (respecting deep nesting), or the current submodule if it is
a `main` submodule (so if you hit `racket-run-main` when you're in a
`main` module, the command does what you'd expect).

This also adds the "C-c C-m" (or "C-c RET") key binding for
`racket-run-main`.

This commit also changes `racket-test` to run the `test` submodule of
the current module (so you can run it within deeply nested
submodules), or the current module if it is a `test` submodule.


Reference.md
```diff
@@ -329,6 +342,8 @@ buffer are Emacs buttons -- which you may navigate among using
 TAB, and activate using RET -- for [`racket-visit-definition`](#racket-visit-definition) and
 [`racket-doc`](#racket-doc).
 
+(fn &optional PREFIX)
```
lenary: I ran `make doc` and this appeared, which I'm not sure is intended. But I'm leaving it as this file is auto-generated.

lenary: The new command I added is there. I’m just confused that this appeared as well.

greghendershott:
Thank you for submitting a pull request!

I'm not entirely clear on the intended use -- especially for deeply nested `main` or `test` submodules. The thing about `main` and `test` as submodules of the _topmost_ module, is that Racket treats them specially. `racket` will run the topmost `main` submodule, and `raco test` will run the topmost `test` submodule.

But not (AFAIK) do anything with any more-deeply nested. So, I'm not grokking why to do that. Why name the deeply nested module `main` instead of `bob`?  And so then also why the special support in racket-mode? Can you say more about this?

---

As for the topmost (not-deeply-nested) ones:

- In racket-mode, `racket-test` was a special command from the start. Later, I added the general ability to put point in any submodule and just run it. That _almost_ makes `racket-test` unnecessary. But `racket-test` can also enable coverage. And, it is handy to kick off tests without moving point.

- As for `main`, I usually have it at the very top or bottom of a file named `main.rkt`. So it hasn't bugged me too much. But I can see the convenience.

---

p.s. A super tiny point: Using up <kbd>C-c RET</kbd> for this seems a shame. But I agree <kbd>C-c C-m</kbd> would be mnemonic. And hey it's Emacs so anyone can rebind any keys; these are all just defaults. :smile:

greghendershott:
I realize that was pretty long-winded. I guess my TL;DR would be: What if we were to simply add `racket-run-main` a.k.a. <kbd>C-c C-m</kbd> to run `main` submodules of the topmost module, only? And `racket-test` doesn't change in this regard?

lenary:
I'm probably weird in that I sometimes structure modules that have an inner submodule, with an (inefficient) copy of the functionality of the containing module. In this case, i find it useful to add a test submodule to the inner module to check functionality, and then the outer module's test module just compares itself to the inner module. The reason I often have multiple implementations of the same functionality is to do with my work, but I'd like to have a way to run the innermost test module in a module like so:

```
#lang racket

;; <outer functionality>

(module inner racket
  ;; <inner functionality>
  (module+ test
    ;; <inner functionality tests>
    ))

(module+ test
  (require (submod ".." inner))
  ;; <outer functionality tests, asserting equivalence with inner functionality>
  )
```

In this case, I'd like to be able to run `(submod <file.rkt> inner test)` if I'm inside `inner` rather than the `test` module in the outermost module.

I realise `raco test` can't run these, but I had been managing by using `C-c C-c` with my point inside the inner test module - and was wanting a way with my point inside the inner module, so I didn't have to move to set off tests. I think of `C-c C-t` running the 'test' submodule of the current module, or the current module if it's already a `test` submodule.

I realise I also have a bug in the checking code, which I am attempting to fix, and will push a fix to in a bit.

Reference.md
lenary: Ok I've attempted to debug this and I'm not clever enough. I'll let you re-run `make doc` and see if that removes the line.

greghendershott: > The new command I added is there.
Ah! I misunderstood the diff snippet. (I've forgotten to add a command to that list. So I just assumed, without looking more carefully.)

greghendershott:
Thanks for explaining how you work, that helps me understand.

- Re `racket-test` and "deeper" `test` submodules: What happens if you <kbd>C-u C-c C-t</kbd> to run with coverage?  Does it work as expected?
    - If yes: Great.
    - If not: I'm not eager to futz with making coverage work in this scenario -- and keep working forever. Maybe it should be documented not to work? Maybe even disabled (e.g. '(when coverage ___)` becomes "when coverage and not a 'deeper' test submodule"?

- Just to spitball (especially if `racket-test` coverage won't work): Is there some simpler, single command we could add instead?  Something like, "run the next submodule after point, regardless of its name (`main`, `test`, or `bob`, whatever)"?  Would that work well enough we could simply do that, instead?

lenary:
In answer to the "coverage" issue, I believe coverage does work.

I ran `C-u C-c C-t` in a submodule, and in that submodule's test module, both times, I got the following in the `*Messages*` buffer:

```
Running tests with coverage instrumentation enabled...
Processing coverage results...
Full coverage.
```

I'm not sure if "Full Coverage" is correct for those tests, and I've never used coverage before, so don't know what the output of this command should be.

greghendershott:
It doesn't seem to work for me. With this file:

```racket
#lang racket/base

(module inner racket/base
  (define (f x) x)
  (module+ test
    (require rackunit)
    (check-equal? (f 0) 0)))

(define (f x) x)
(module+ test
  (require rackunit)
  (check-equal? (f 0) 0))
```

If I move to beginning of file and <kbd>C-u C-c C-t</kbd>, it colors both expressions of `inner` in red and says `Missing coverage in 2 place(s).`

screen shot 2018-05-27 at 9 54 45 am

In other words, the coverage stuff has no awareness of non-module-level `test` submodules. Attempting to make it handle this feels like diminishing returns.

lenary:
Thanks for taking the time on this, but I'm going to kill this PR. 

Am I right in thinking that `racket-racket` will invoke `racket <filename>` in a shell somehow? That should run the `main` module IIUC, so I guess that's what I should be binding to `C-c C-m` in my own emacs config.

PR #308, opened 2018-06-29
Improve racket--version for speed

xuchunyang:
`racket --version` is the fastest way to get the version

```
~ $ time racket -e '(version)'
"6.12"
        0.79 real         0.43 user         0.19 sys
~ $ time racket -l racket/base -e '(version)'
"6.12"
        0.09 real         0.06 user         0.02 sys
~ $ time racket --version
Welcome to Racket v6.12.
        0.02 real         0.01 user         0.01 sys
```

### DIFF
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -147,13 +147,12 @@ Never changes selected window."
   "Get the `racket-program' version as a string."
   (with-temp-message "Checking Racket version..."
     (with-temp-buffer
-      (call-process racket-program
-                    nil                  ;infile: none
-                    t                    ;destination: current-buffer
-                    nil                  ;redisplay: no
-                    "-e"
-                    "(version)")
-      (eval (read (buffer-substring (point-min) (point-max)))))))
+      (call-process racket-program nil t nil "--version")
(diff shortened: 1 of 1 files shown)


greghendershott:
Thanks!

That's a significant speed-up -- especially from the status quo `racket -e '(version)'` to `racket -l racket/base -e '(version)'`.

The third one using `racket --version` makes me hesitate, because it seems _slightly_ "brittle" to regex match the welcome string? But I guess it's OK; probably Racket wouldn't change that someday, because it would break other things, too.

---

Question: If we're going to use regex, should we use [`save-match-data`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Saving-Match-Data.html)? That is, change this:

```elisp
(re-search-forward "[0-9]+\\(?:\\.[0-9]+\\)*")
(match-string 0)
```

to this:

```elisp
(save-match-data
  (re-search-forward "[0-9]+\\(?:\\.[0-9]+\\)*")
  (match-string 0))
```

greghendershott:
Merged -- thanks!

PR #309, opened 2018-07-04
Ignore empty requires for racket-tidy-requires

xuchunyang:
`racket-tidy-requires` and `racket-trim-requires` don't handle buffer without any require very well, for example

    #lang racket
    
    (displayln "Hello world")

then running `racket-tidy-requires` will raise such error 

    Debugger entered--Lisp error: (wrong-type-argument integer-or-marker-p nil)
      goto-char(nil)
      (if (string-equal "" new) nil (goto-char (racket--top-level-requires 'kill)) (insert (concat new "\n")))
      (let* ((reqs (racket--top-level-requires 'find)) (new (and reqs (racket--repl-command "requires/tidy %S" reqs)))) (if (string-equal "" new) nil (goto-char (racket--top-level-requires 'kill)) (insert (concat new "\n"))))
      racket-tidy-requires()
      funcall-interactively(racket-tidy-requires)
      call-interactively(racket-tidy-requires nil nil)
      command-execute(racket-tidy-requires)

I think in this case it should simply pass without such error, so I made the pr.

### DIFF
--- a/racket-edit.el
+++ b/racket-edit.el
@@ -477,12 +477,12 @@ file using `#lang`. It does *not* work for `require`s inside
 
 See also: `racket-trim-requires' and `racket-base-requires'."
   (interactive)
-  (let* ((reqs (racket--top-level-requires 'find))
-         (new (and reqs
-                   (racket--repl-command "requires/tidy %S" reqs))))
-    (unless (string-equal "" new)
-      (goto-char (racket--top-level-requires 'kill))
-      (insert (concat new "\n")))))
+  (let ((reqs (racket--top-level-requires 'find)))
+    (when reqs
+      (let ((new (racket--repl-command "requires/tidy %S" reqs)))
+        (unless (string-equal "" new)
+          (goto-char (racket--top-level-requires 'kill))
+          (insert (concat new "\n")))))))
 
 (defun racket-trim-requires ()
   "Like `racket-tidy-requires' but also deletes unused modules.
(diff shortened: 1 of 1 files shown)


greghendershott:
Good point.

Do you think it might be simpler to change `(racket--top-level-requires 'find)` to raise a `user-error`?

Maybe something like this:

```diff
modified   racket-edit.el
@@ -563,7 +563,8 @@ such as changing `#lang typed/racket` to `#lang typed/racket/base`."
 (defun racket--top-level-requires (what)
   "Identify all top-level requires and do WHAT.
 
-When WHAT is 'find, returns the top-level require forms.
+When WHAT is 'find, returns the top-level require forms. If there
+aren't any, does `user-error'.
 
 When WHAT is 'kill, kill the top-level requires, returning the
 location of the first one."
@@ -581,7 +582,10 @@ location of the first one."
           (when (eq 'kill what)
             (kill-sexp -1)
             (delete-blank-lines))))
-      (if (eq 'kill what) first-beg requires))))
+      (if (eq 'kill what)
+          first-beg
+        (or requires
+            (user-error "No top-level requires"))))))
```

That's just a quick sketch, to see what you think; I didn't try it.

xuchunyang:
I'm not sure all use cases of `(racket--top-level-requires 'find)` (Maybe the caller wants to do something when it returns `nil`), thus I can't decide.

greghendershott:
You're right. Actually `racket-base-requires` also calls it and needs the `nil` return value.

So the approach of your PR is better.

greghendershott:
Thanks again for submitting this PR!

I'm closing it without merging only because I happened to fix the same issue in commit e00cb29 while working on a related problem.

PR #337, opened 2018-10-03
don't cache the default browser

### LINKED ISSUE #336: The default browser setting on macOS is cached in Racket Mode
Hi Greg,

The browser to open document is cached [here](https://github.com/greghendershott/racket-mode/blob/master/racket/commands/help.rkt#L120). if the system-wide default browser is changed, Racket mode will keep using the old one, which seems a little surprising and inconsistent with other apps/mode/command. 

I wonder if we can call the function mac-default-brower every time. If it is feasible, I will send a pull request. Or are there any performance concerns?

capfredf:
Fix #336

### DIFF
--- a/racket/commands/help.rkt
+++ b/racket/commands/help.rkt
@@ -90,11 +90,10 @@
 (require 'mac-default-browser
          'browse-file-url/mac)
 
-(define mac-browser (mac-default-browser))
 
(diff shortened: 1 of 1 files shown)


greghendershott:
Thanks!

PR #339, opened 2018-10-09
Improve `{beginning,end}-of-defun` and `add-log-current-defun-function`

alex-hhh:
The back end functions required to move forward and backward between functions have been updated to recognize nested constructs, such as method definitions inside classes or inner definitions in functions.

For the example below, `C-M-a` command will move from "start point 1" in the following order:

* inner-method2
* method2
* method1
* member-var1
* my-class%
* function1
* var1
* beginning of buffer

From "start-point2" it will navigate only the toplevel definitions and will not go inside the class methods.

I have also added a function to determine the current function name for the `C-x 4 a` command, and this function is also aware of inner definitions, for example at "start point 1", the function name would be "my-class% method2 inner-method2"

```racket
#lang racket

(define var1 'var1)
(define (function1) (void))

(define my-class%
  (class object%
    (init) (super-new)
    (define member-var1 'member-var1)
    (define/public (method1) (void))
    (define/public (method2)
      (define (inner-metod2)
        ;; example start point 1
        (void))
      (void))))

(define (function2)
  (define (inner-function2)
    ;; example start point 2
    (void))
  void)
```

### DIFF
--- a/racket-common.el
+++ b/racket-common.el
@@ -239,7 +239,9 @@ property whose value is STRING. The close | syntax is set by
   (setq-local outline-regexp ";;; \\|(....")
   (setq-local completion-at-point-functions (list #'racket-complete-at-point))
   (setq-local eldoc-documentation-function nil)
-  (setq-local beginning-of-defun-function #'racket--beginning-of-defun-function))
+  (setq-local beginning-of-defun-function #'racket--beginning-of-defun-function)
+  (setq-local end-of-defun-function #'racket--end-of-defun-function)
+  (setq-local add-log-current-defun-function #'racket--current-defun-name))
 
 
 ;;; Insert lambda char (like DrRacket)
@@ -628,14 +630,77 @@ This function is a suitable element for the list variable
 
 ;;; racket--beginning-of-defun
 
+(defconst racket--defun-start-rx "(define\\(/[^ \t\n]*\\)?[ \t\n]*(?\\(\\(:?\\sw\\|\\s_\\)+\\)"
+  "Regexp matching the start of a definition.
+The defined name is the second regexp group.")
+
 (defun racket--beginning-of-defun-function ()
-  "Like `beginning-of-defun' but aware of Racket module forms."
+  "Move point backward to the beginning of a definition.
+The code moves only to the next inner most definition, if nested
+defines are present. If the point is already on a definition
+which is not a toplevel one, it will move to the next outer
+definition, thus repeated calls of this function will move up on
+the nested definitions."
   (let ((orig (point)))
-    (racket--escape-string-or-comment)
-    (pcase (racket--module-level-form-start)
-      (`() (ignore-errors (backward-sexp 1)))
-      (pos (goto-char pos)))
+    (catch 'found
+      ;; If we are after the close paren of a definition, but before
+      ;; the next definition, move to the definition start...
+      (ignore-errors (backward-sexp))
+      (when (looking-at racket--defun-start-rx)
+        (throw 'found (point)))
+      (goto-char orig)
+      (let ((parens (reverse (nth 9 (syntax-ppss (point))))))
+        ;; If we are inside a nested sexp, try to move to the
+        ;; innermost definition...
+        (dolist (pos parens)
+          (goto-char pos)
+          (when (looking-at racket--defun-start-rx)
+            (throw 'found pos)))
+        ;; ... we could not find a definition start, move to top level
+        ;; and start scanning sexps backwards until we find a
+        ;; definition, or reach the beginning of the buffer.
+        (unless (eq parens nil)
+          (goto-char (car (last parens))))
+        (backward-sexp 1)
+        (while (not (or (bobp) (looking-at racket--defun-start-rx)))
+          (backward-sexp 1))))
+    (goto-char (or (racket--sexp-comment-start (point)) (point)))
     (/= orig (point))))
+
+(defun racket--end-of-defun-function ()
+  "Move point to the end of the current function definition.
+This is intended to be used as the `end-of-defun-function' value,
+and assumes point is set up correctly."
+  (let ((orig (point)))
+    (condition-case err
+        (forward-sexp 1)
+      (scan-error
+       ;; scan errors happen if we are inside a nested definition and
+       ;; move to the end.  Go outside the nested definition in this
+       ;; case.
+       (let ((parens (nth 9 (syntax-ppss orig))))
+         (if (eq parens nil)
+             (goto-char (point-max))
+           (progn
+             (goto-char (car parens))
+             (forward-sexp 1))))))))
+
+(defun racket--current-defun-name ()
+  "Return the current function name.
+
+If the point is inside a nested function definition, return a
+string that represents the concatenation of the nested function names."
+  (save-excursion
+    (unless (looking-at racket--defun-start-rx)
+      (racket--beginning-of-defun-function))
+    (when (looking-at racket--defun-start-rx)
+      (let ((name (list (match-string-no-properties 2)))
+            (parens (reverse (nth 9 (syntax-ppss (point))))))
+        (dolist (pos parens)
+          (goto-char pos)
+          (when (looking-at racket--defun-start-rx)
+            (push (match-string-no-properties 2) name)))
+        (string-join name " ")))))
 
 (defun racket--module-level-form-start ()
   "Start position of the module-level form point is within.
@@ -674,7 +739,7 @@ Allows #; to be followed by zero or more space or newline chars."
     (goto-char pos)
     (while (memq (char-before) '(32 ?\n))
       (goto-char (1- (point))))
-    (when (string= "#;" (buffer-substring-no-properties (- (point) 2) (point)))
+    (when (string= "#;" (buffer-substring-no-properties (max 1 (- (point) 2)) (point)))
       (- (point) 2))))
 
 
--- a/racket-common.el
+++ b/racket-common.el
@@ -26,6 +26,7 @@
 (require 'racket-indent)
 (require 'racket-ppss)
 (require 'racket-util)
+(require 'subr-x)
 
 (declare-function racket-complete-at-point "racket-complete.el" (&optional predicate))
 
--- a/racket-common.el
+++ b/racket-common.el
@@ -634,6 +634,21 @@ This function is a suitable element for the list variable
   "Regexp matching the start of a definition.
 The defined name is the second regexp group.")
 
+(defun racket--collect-open-paren-positions (pos)
+  "Return a list containing the positions of open parens for the point at POS.
+Outermost paren is first in the list"
+  ;; NOTE: Emacs 26.1 returns this the list of open parens in the 9th
+  ;; position of the `syntax-ppss' return value, however this
+  ;; functionality is new in 26.1 and and older Emacs versions return
+  ;; opaque state information in that position
+  (let ((parens '())
+        (ppss (syntax-ppss pos)))
+    (while (racket--ppss-containing-sexp ppss)
+      (let ((paren-pos (racket--ppss-containing-sexp ppss)))
+        (push paren-pos parens)
+        (setq ppss (syntax-ppss paren-pos))))
+    parens))
+
 (defun racket--beginning-of-defun-function ()
   "Move point backward to the beginning of a definition.
 The code moves only to the next inner most definition, if nested
@@ -649,7 +664,7 @@ the nested definitions."
       (when (looking-at racket--defun-start-rx)
         (throw 'found (point)))
       (goto-char orig)
-      (let ((parens (reverse (racket--ppss-parens (syntax-ppss (point))))))
+      (let ((parens (reverse (racket--collect-open-paren-positions (point)))))
         ;; If we are inside a nested sexp, try to move to the
         ;; innermost definition...
         (dolist (pos parens)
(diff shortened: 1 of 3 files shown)


alex-hhh:
Looks like the build fails, because Emacs 24.3 does not have `subr-x` and `string-join`.

racket-common.el
```diff
@@ -628,15 +631,78 @@ This function is a suitable element for the list variable
...
+    (unless (looking-at racket--defun-start-rx)
+      (racket--beginning-of-defun-function))
+    (when (looking-at racket--defun-start-rx)
+      (let ((name (list (match-string-no-properties 2)))
+            (parens (reverse (nth 9 (syntax-ppss (point))))))
+        (dolist (pos parens)
+          (goto-char pos)
+          (when (looking-at racket--defun-start-rx)
+            (push (match-string-no-properties 2) name)))
+        (string-join name " ")))))
```
greghendershott: If you closed this PR only because Emacs 24.3 lacks `string-join` -- the definition of `string-join` in 26.1 seems to be simply `(mapconcat 'identity strings separator)`. So would something like `(mapconcat #'identity name " ")` work here?

racket-common.el
```diff
@@ -628,15 +631,78 @@ This function is a suitable element for the list variable
...
+  "Return the current function name.
+
+If the point is inside a nested function definition, return a
+string that represents the concatenation of the nested function names."
+  (save-excursion
+    (unless (looking-at racket--defun-start-rx)
+      (racket--beginning-of-defun-function))
+    (when (looking-at racket--defun-start-rx)
+      (let ((name (list (match-string-no-properties 2)))
+            (parens (reverse (nth 9 (syntax-ppss (point))))))
```
greghendershott: To make code a little more self-documenting, I added `racket-ppss.el` with accessor aliases. But it only goes up to `(nth 8)`. Perhaps you could add one for `(nth 9)` named something like `racket--ppss-open-paren-positions`, and use it here? (If that seems like too much busywork for you, no worries; I might come back and do it later myself.)

alex-hhh:
It looks like I pressed the wrong button when I added my comment, as I was trying to do too many things at once last night.  I will address your comments tonight (it is early morning now here).

I also wrote a flymake integration for Racket which works better than the flycheck one  (in particular, it can report bad requires and errors in other racket files than the current one).  You can find it below, but this code might also work only in recent Emacs versions.  Are you interested in merging that in? 

https://github.com/alex-hhh/racket-mode/tree/ah/racket-flymake

greghendershott:
> It looks like I pressed the wrong button when I added my comment, as I was trying to do too many things at once last night. I will address your comments tonight (it is early morning now here).

I thought maybe you decided to close the PR when you discovered it wouldn't work in Emacs 24.3. I just wanted to point out that it might be easy (-ish?) to make it work. If you want to try, that's great! But if you don't, that's fine, too, I didn't mean to be pushy!

> I also wrote a flymake integration for Racket which works better than the flycheck one (in particular, it can report bad requires and errors in other racket files than the current one). You can find it below, but this code might also work only in recent Emacs versions. Are you interested in merging that in?

I recently did a PR to MELPA for moving the `.rkt` files down under `racket/`. In the queue I thought I noticed a new flymake-racket package... let me check... yes it's [here](https://melpa.org/#/flymake-racket).

I'm vaguely aware of some backstory, like, at first there was flymake, then flycheck came out, then the original flycheck author decided to step away from it and emacs because they didn't like emacs-devel culture, then flymake got some new attention, but only starting in version 26... maybe??? I haven't kept up with it Maybe there's a good reason for this to be a separate package; and it does now exist? 

But I don't have a strong opinion either way, so please let me know if there's a good reason to merge something into racket-mode vs. suggest people install/use flymake-racket?

alex-hhh:
Regarding flymake support, I prefer packages that provide a comprehensive set of features. For example, when I install `racket-mode`, I expect Emacs to support editing Racket programs: this includes syntax highlighting, indentation, code navigation, a REPL and, yes, flymake support. I think this is more convenient than having to install several packages, each providing just a small portion of the functionality: I have to discover, install and configure each such package independently.

However, given the large number of such small packages on MELPA, I think not many people share my views.

----

As a side note, and before others point it out :-), I also maintain a separate `sql-indent` package which only deals with SQL code indentation, but this is only because the Emacs maintainers did not want to add this functionality to the built-in sql.el, instead they wanted a separate GNU ELPA package.

capfredf:
On a possible relate note: flymake-racket could use a different executable from one Racket mode uses, which may confuse users. 

In general, I think an all-in-one solution provided by a specified language mode usually offers a better user experience

greghendershott:
@alex-hhh Hmm -- one thing just occurred to me: Do we know if `(elt (syntax-ppss) 9)` worked back in Emacs 24.3?

I'm worried because I don't remember why I didn't originally define an accessor like`racket--syntax-parens` for elt 9. Was I just lazy and stopped at 8? Or, was it because it wasn't documented and/or implemented back then?

If you have a chance to try with Emacs 24.3 and let me know, that would be great. Otherwise, I'll check myself when I have a chance.

greghendershott:
Thanks for change to remain compatible with 24.3. I haen't fetched that yet, but I'd fetched through your previous commits and was using it live here.

I just encountered the following. In a racket-mode buffer, type:

```
;;; If we (might)
```

When I try to type that closing paren, with debug errors toggled on, I get:

```
Debugger entered--Lisp error: (scan-error "Containing expression ends prematurely" 30 30)
  scan-sexps(31 -1)
  forward-sexp(-1)
  backward-sexp(1)
  (while (not (or (bobp) (looking-at racket--defun-start-rx))) (backward-sexp 1))
  (let ((parens (reverse (racket--ppss-parens (syntax-ppss (point)))))) (let ((--dolist-tail-- parens) pos) (while --dolist-tail-- (setq pos (car --dolist-tail--)) (goto-char pos) (if (looking-at racket--defun-start-rx) (progn (throw 'found pos))) (setq --dolist-tail-- (cdr --dolist-tail--)))) (if (eq parens nil) nil (goto-char (car (last parens)))) (backward-sexp 1) (while (not (or (bobp) (looking-at racket--defun-start-rx))) (backward-sexp 1)))
  (catch 'found (condition-case nil (progn (backward-sexp)) (error nil)) (if (looking-at racket--defun-start-rx) (progn (throw 'found (point)))) (goto-char orig) (let ((parens (reverse (racket--ppss-parens (syntax-ppss (point)))))) (let ((--dolist-tail-- parens) pos) (while --dolist-tail-- (setq pos (car --dolist-tail--)) (goto-char pos) (if (looking-at racket--defun-start-rx) (progn (throw 'found pos))) (setq --dolist-tail-- (cdr --dolist-tail--)))) (if (eq parens nil) nil (goto-char (car (last parens)))) (backward-sexp 1) (while (not (or (bobp) (looking-at racket--defun-start-rx))) (backward-sexp 1))))
  (let ((orig (point))) (catch 'found (condition-case nil (progn (backward-sexp)) (error nil)) (if (looking-at racket--defun-start-rx) (progn (throw 'found (point)))) (goto-char orig) (let ((parens (reverse (racket--ppss-parens (syntax-ppss (point)))))) (let ((--dolist-tail-- parens) pos) (while --dolist-tail-- (setq pos (car --dolist-tail--)) (goto-char pos) (if (looking-at racket--defun-start-rx) (progn (throw 'found pos))) (setq --dolist-tail-- (cdr --dolist-tail--)))) (if (eq parens nil) nil (goto-char (car (last parens)))) (backward-sexp 1) (while (not (or (bobp) (looking-at racket--defun-start-rx))) (backward-sexp 1)))) (goto-char (or (racket--sexp-comment-start (point)) (point))) (/= orig (point)))
  racket--beginning-of-defun-function()
  beginning-of-defun-raw(nil)
  beginning-of-defun()
  paredit-current-parse-state()
  paredit-in-string-p()
  paredit-move-past-close-and(41 #f(compiled-function () #<bytecode 0x4118bbbd>))
  paredit-move-past-close(41)
  paredit-close-round()
  funcall-interactively(paredit-close-round)
  call-interactively(paredit-close-round nil nil)
  command-execute(paredit-close-round)
```

greghendershott:
Another recipe for the same bug:

- Enable `paredit-mode` in `racket-repl-mode` buffer.
- Go to `racket-repl-mode`.
- At a prompt, type `1`.
- Type <kbd>BACKSPACE</kbd>.

```
Debugger entered--Lisp error: (scan-error "Unbalanced parentheses" 33255 1)
  scan-sexps(33257 -1)
  forward-sexp(-1)
  backward-sexp(1)
  (while (not (or (bobp) (looking-at racket--defun-start-rx))) (backward-sexp 1))
  (let ((parens (reverse (racket--ppss-parens (syntax-ppss (point)))))) (let ((--dolist-tail-- parens) pos) (while --dolist-tail-- (setq pos (car --dolist-tail--)) (goto-char pos) (if (looking-at racket--defun-start-rx) (progn (throw 'found pos))) (setq --dolist-tail-- (cdr --dolist-tail--)))) (if (eq parens nil) nil (goto-char (car (last parens)))) (backward-sexp 1) (while (not (or (bobp) (looking-at racket--defun-start-rx))) (backward-sexp 1)))
  (catch 'found (condition-case nil (progn (backward-sexp)) (error nil)) (if (looking-at racket--defun-start-rx) (progn (throw 'found (point)))) (goto-char orig) (let ((parens (reverse (racket--ppss-parens (syntax-ppss (point)))))) (let ((--dolist-tail-- parens) pos) (while --dolist-tail-- (setq pos (car --dolist-tail--)) (goto-char pos) (if (looking-at racket--defun-start-rx) (progn (throw 'found pos))) (setq --dolist-tail-- (cdr --dolist-tail--)))) (if (eq parens nil) nil (goto-char (car (last parens)))) (backward-sexp 1) (while (not (or (bobp) (looking-at racket--defun-start-rx))) (backward-sexp 1))))
  (let ((orig (point))) (catch 'found (condition-case nil (progn (backward-sexp)) (error nil)) (if (looking-at racket--defun-start-rx) (progn (throw 'found (point)))) (goto-char orig) (let ((parens (reverse (racket--ppss-parens (syntax-ppss (point)))))) (let ((--dolist-tail-- parens) pos) (while --dolist-tail-- (setq pos (car --dolist-tail--)) (goto-char pos) (if (looking-at racket--defun-start-rx) (progn (throw 'found pos))) (setq --dolist-tail-- (cdr --dolist-tail--)))) (if (eq parens nil) nil (goto-char (car (last parens)))) (backward-sexp 1) (while (not (or (bobp) (looking-at racket--defun-start-rx))) (backward-sexp 1)))) (goto-char (or (racket--sexp-comment-start (point)) (point))) (/= orig (point)))
  racket--beginning-of-defun-function()
  beginning-of-defun-raw(nil)
  beginning-of-defun()
  paredit-current-parse-state()
  paredit-in-string-p()
  paredit-backward-delete(nil)
  funcall-interactively(paredit-backward-delete nil)
  call-interactively(paredit-backward-delete nil nil)
  command-execute(paredit-backward-delete)
```

alex-hhh:
I will fix the error -- there was a bug in how the code parses an unbalanced paren inside a comment.

alex-hhh:
The programming mode itself defines what a function is and where it begins, this is true even in C or Python which don't use nested parens.  The rest of the Emacs facilities build on top of this definition: for example, `narrow-to-defun` narrows to the region between `beginning-of-defun` and `end-of-defun`.  Emacs itself has no other opinion of what a function is in a particular programming language.

This pull request changes the racket mode to consider "function definitions" things like inner functions, class methods, class definitions and things like `(define f (lambda (x) x))` -- I found this definition mode useful when working with code that uses classes a lot.  Yes, the code also stops at plain defines like `(define x 1)`, but fixing that requires a lot more effort and I personally actually like it this way: it stops at member variable definitions inside classes too.

Also, the previous `b-o-d` implementation in Racket used to stop at all top level forms, for example, if you have a `(display "Hello")` at top level, `C-M-a` would stop there, and that form is not any kind of definition.

Binding another function to `C-M-a` is not the right thing to do for two reasons: first, `beginning-of-defun` does more than just move the point to the function definition (this is why we provide a value for `beginning-of-defun-function`, not replace `beginning-of-defun`) and second, other Emacs facilities build on top of this, like `narrow-to-defun`.

---

On a related note, it looks like `racket-send-definition` is using
`beginning-of-defun`, `end-of-defun`, I will change them to determine the top-level form and send that.  And even the Emacs Lisp implementation of that function, while having `defun` in its name, does not actually evaluate only function definitions, this is from its help string:

```
(edebug-eval-defun EDEBUG-IT)

Evaluate the top-level form containing point, or after point.
```

-----

The entire basis for this pull request is to implement `b-o-d` as defined above, and If you disagree with this definition, this pull request is the wrong approach.  Can you let me know if you disagree with this definition, as in that case there is no point on working on this anymore.  Thanks, Alex.

greghendershott:
It's not that I disagree with the approach. It's that I'm juggling too many other things lately. So just trying to keep up with the PR by asking questions. I haven't had time to think about it very much.

- I do like the idea of it handling nested function definitions.

- As for non-function defines, I asked, but I like your answer and I think it's OK or even desirable for it to visit those, too.

- All that sounds great as a new behavior -- for human users of racket-mode users e.g. pressing <kbd>C-M-a</kbd>. The thing that makes me nervous is the new behavior for other things using `beginning-of-defun`. For instance `paredit` does, as in the call stack above. If the behavior changes, will it cause problems? I don't know. I'm asking. (I do understand the little bug I found, sounds like just a small mistake in the implementation. Not arising from the change in behavior per se. But are there other bugs waiting? If you have time to think about it awhile and feel good about it -- great.)

So again, I really appreciate the PR. I don't have time to review it as carefully as I'd like. So I'm asking you some of the questions I'd ask, if I had more time.

Thanks!

alex-hhh:
I cannot offer you any guarantees that the code has no bugs or that it will not conflict with other packages -- there are too many packages out there for such a verification to be practical.  However:

* I will try and fix any problems or defects that are bought to my attention.  I think I showed that with this pull request and I am happy to do that even after (if?) this code is merged
* You can always `git revert` the changes, if it turns out that these changes create a big problem with Emacs or some other package and we find no other way to address these problems

I don't think that it is a good idea to have a difference between a user level `beginning-of-defun` and a program level one, as this would only create confusion when the user can see one beginning of defun and a program will use another.

Issue #359, opened 2019-02-07
Company mode freezes Emacs after run-racket


hencq:
After updating racket-mode, company-mode started freezing Emacs. I noticed this only happens after I run run-racket. Before that it works fine. Disabling company-mode fixes it.

After digging around a little bit, it seems that it happens due to a call to

    (racket--cmd/await `(path+md5))

which gives a "racket-command process timeout" error. The problem goes away when in command-server.rkt I change
     
    [`(path+md5)  (cons (or path 'top) md5)]

to

    [`(path+md5) (cons 'top "")                                     ]

I tried digging around a little bit more in there, but got stuck. Happy to provide more details though if that's useful.

See below for debugging details:
<details>
```
((emacs-version "26.1")
 (emacs-uptime "52 minutes, 36 seconds")
 (system-type darwin)
 (major-mode racket-mode)
 (racket--el-source-dir "/Users/usernam/.emacs.d/elpa/racket-mode-20190206.134/")
 (racket--rkt-source-dir "/Users/username/.emacs.d/elpa/racket-mode-20190206.134/racket/")
 (racket-program "racket")
 (racket-command-port 55555)
 (racket-command-timeout 10)
 (racket-memory-limit 2048)
 (racket-error-context medium)
 (racket-retry-as-skeleton t)
 ...)
(enabled-minor-modes
 (auto-composition-mode)
 (auto-compression-mode)
 (auto-encryption-mode)
 (auto-fill-mode)
 (auto-save-mode)
 (blink-cursor-mode)
 (column-number-mode)
 (company-mode)
 (diff-auto-refine-mode)
 (electric-indent-mode)
 (file-name-shadow-mode)
 ...)
(disabled-minor-modes
 (abbrev-mode)
 (auto-complete-mode)
 (auto-fill-function)
 (auto-save-visited-mode)
 (beancount-mode)
 (buffer-read-only)
 (cl-old-struct-compat-mode)
 (company-search-mode)
 (compilation-in-progress)
 (compilation-minor-mode)
 (compilation-shell-minor-mode)
 ...)
```
</details>

hencq:
Digging into it some more, I saw the last commit touched paths as well. I think the issue is in

   (define (path->string/emacs p)
       (match (map path->string (explode-path p))
        [(list* (pregexp "([a-zA-z]:)\\\\" (list _ drive)) vs)
        #:when (eq? 'windows (system-type))
         (string-join (cons drive vs))]
        [_ p]))

[...]

PR #360, opened 2019-02-08
Fix for Emacs freezing with company mode #359 introduced by f7d37a5

hencq:
path->string/emacs returns a path if it was passed a path. This fix makes it return a string.


greghendershott:
Thank you!

Issue #300, opened 2018-04-17
Function for sending string to repl


EliasCarl:
Hi, first of all thanks for a great package.

As a complete beginner to elisp (and racket) I was trying to extend the following function in lispy-mode https://github.com/abo-abo/lispy/blob/58a3f21a1946768bb6d20a6694d37d749cb136ce/lispy.el#L6429. I could get it to work by copying racket--send-region-to-repl and modify it to take a string instead and give that function to lispy--eval. Just wondering if there's already such a "send-string-to-repl" function to use instead of my ugly copy & paste solution.

greghendershott:
Hi! I glanced at this when you submitted, saw it didn't seem to be an urgent bug, and... forgot to loop back and make some reply. Sorry!

The copy/paste is the least-worst available solution.

You could almost just do `(comint-send-string (racket--get-repl-buffer-process) "(+ 1 1)\n")` -- the `\n` is important -- if you don't care about details like checking for a REPL and starting one if necessary.

xuchunyang:
> I could get it to work by copying racket--send-region-to-repl and modify it to take a string instead and give that function to lispy--eval.

That's not enough, you still need to find a way to capture the return value, which is required by lispy--eval, but not provided by racket--send-region-to-repl.

greghendershott:
Ah, OK. I don't use lispy myself and unfortunately currently don't have time to learn about it.

I did take a very quick look at its repo. I notice that its support for e.g. geiser uses some "private" functions: `geiser--`_name_.[^1] This seems non-ideal.

So, if anyone wanted to make a PR to lispy to add racket-mode support, I do have a suggestion: First make a PR here to add a `racket-lispy-eval` public "API" function, and have lispy use _that_. 

(Of course the implementation of `racket-lispy-eval` may use whatever private racket-mode functions it needs.)

---

[^1]: Emacs Lisp doesn't have modules or namespaces, so, `mode--`_name_ is a convention meaning "private" as opposed to `mode-`_name_ meaning "public". Two vs. one hyphen.

xuchunyang:
I don't use lispy either. But I am interested because I would like racket-mode's `C-x C-e` to be able to show the result in echo area, just like `C-x C-e` in emacs-lisp-mode.

BTW, for python, there is a `python-shell-send-string-no-output`

```
(python-shell-send-string-no-output "2**10\n")
;; => "1024"
```

greghendershott:
I see.

[...]

Issue #344, opened 2018-11-24
REPL periodically freezes Emacs


DarrenN:
## Background

I wanted to start this issue so I can being gathering more information.

Periodically when I have the Racket REPL open in `racket-mode` Emacs will freeze and peg a core at 100%. I cannot determine when it will happen. Sometimes I will step away from my laptop for a moment and come back and its hung, other times I will be in the middle of editing code and it will freeze. I'm not 100% sure its related to the REPL, but it only happens when I have the REPL open. I have to force quit Emacs via Activity Moniotor when this happens. Sometimes its gets really bad and I revert to running code on the CLI instead of using the REPL.

The code I'm working on it isn't doing anything fancy with threads that would block the REPL (and even when I do I can break out with C-c).

What other information can I grab? When it does lock I can sample the running process from OSX's Activity Monitor.

## Debug information

![screenshot 2018-11-24 09 48 00](https://user-images.githubusercontent.com/215030/48969532-1803be00-efce-11e8-956e-97e5d5ab8c5a.png)

```
((emacs-version "26.1")
 (emacs-uptime "16 hours, 32 minutes, 48 seconds")
 (system-type darwin)
 (major-mode racket-mode)
 (racket--source-dir "/Users/kyushu/.emacs.d/elpa/racket-mode-20180827.1303/")
 (racket-program "racket")
 (racket-memory-limit 2048)
 (racket-error-context medium)
 (racket-history-filter-regexp "\\`\\s *\\S ?\\S ?\\s *\\'")
 (racket-images-inline t)
 (racket-images-keep-last 100)
 (racket-images-system-viewer "display")
 (racket-pretty-print t)
 (racket-indent-curly-as-sequence t)
 (racket-indent-sequence-depth 0)
 (racket-pretty-lambda nil)
 (racket-smart-open-bracket-enable nil))
(enabled-minor-modes
 (async-bytecomp-package-mode)
 (auto-composition-mode)
 (auto-compression-mode)
 (auto-encryption-mode)
 (auto-fill-mode)
 (auto-revert-mode)
 (auto-save-mode)
 (blink-cursor-mode)
 (column-number-mode)
 (company-mode)
 (delete-selection-mode)
 (diff-auto-refine-mode)
 (eldoc-mode)
 (electric-indent-mode)
 (file-name-shadow-mode)
 (font-lock-mode)
 (global-company-mode)
 (global-eldoc-mode)
 (global-font-lock-mode)
 (global-git-commit-mode)
[...]

PR #374, opened 2019-05-11
add some lispy support

odanoburu:
- following instructions in #300, create public racket-mode function
  (`racket-lispy-visit-symbol-definition`) to be used by lispy
  - create helper function `racket--visit-symbol-definition` shared
    between `racket-lispy-visit-symbol-definition` and
    `racket-visit-definition`

- lispy code at https://github.com/abo-abo/lispy/compare/master...odanoburu:racket, I'll submit the PR to lispy after your approval
  - I'll be implementing the lispy functions as I need them, but the two functions in the link are the bare essentials I need

### DIFF
--- a/racket-edit.el
+++ b/racket-edit.el
@@ -202,7 +202,7 @@ Please keep in mind the following limitations:
     (`nil nil)
     (str (racket--visit-symbol-definition str))))
 
-(defsubst racket-lispy-visit-symbol-definition (str)
+(defun racket-lispy-visit-symbol-definition (str)
   "Function called by lispy.el's `lispy-goto-symbol' for Racket
 symbol definition lookup."
   (racket--visit-symbol-definition str))
--- a/racket-edit.el
+++ b/racket-edit.el
@@ -200,14 +200,22 @@ Please keep in mind the following limitations:
   (interactive "P")
   (pcase (racket--symbol-at-point-or-prompt prefix "Visit definition of: ")
     (`nil nil)
-    (str (if (and (eq major-mode 'racket-mode)
-                  (not (equal (racket--repl-file-name+md5)
-                              (cons (racket--buffer-file-name t) (md5 (current-buffer)))))
-                  (y-or-n-p "Run current buffer first? "))
-             (racket--repl-run nil nil
-                               (lambda (_n/a)
-                                 (racket--do-visit-def-or-mod 'def str)))
-           (racket--do-visit-def-or-mod 'def str)))))
+    (str (racket--visit-symbol-definition str))))
+
+(defsubst racket-lispy-visit-symbol-definition (str)
+  "Function called by lispy.el's `lispy-goto-symbol' for Racket
+symbol definition lookup."
+  (racket--visit-symbol-definition str))
+
+(defun racket--visit-symbol-definition (str)
+  (if (and (eq major-mode 'racket-mode)
+           (not (equal (racket--repl-file-name+md5)
+                       (cons (racket--buffer-file-name t) (md5 (current-buffer)))))
+           (y-or-n-p "Run current buffer first? "))
+      (racket--repl-run nil nil
+                        (lambda (_n/a)
+                          (racket--do-visit-def-or-mod 'def str)))
+    (racket--do-visit-def-or-mod 'def str)))
 
 (defun racket-visit-module (&optional prefix)
   "Visit definition of module at point, e.g. net/url or \"file.rkt\".


greghendershott:
Quick glance, this seems good. Just give me a day to find time to review it.

As for #300, although I added a `racket-eval-last-expr`, it shows the value(s) using `message`. Does lispy actually need a wrapper that _returns_ the value(s)?

Also, can lispy handle there being 0 or more Racket values (0 being the `void` case where Racket `print` normally prints nothing)?

I'm just thinking ahead to whether you'll need a similar wrapper for that, like you did here.

greghendershott:
Also, please test the scenario where `racket-repl-mode` doesn't exist or doesn't have a "live" REPL prompt (no Racket back-end process yet)?  Is there a reasonable end user error message?

I changed this recently to (by default) *not* automagically helpfully try to start things. See #344.

odanoburu:
no problem -- take the time you need!

I have only added support for visiting definition and making `RET` work on the REPL. I haven't done anything about the lispy-eval capability, mostly because I don't use it very much, specially when using racket. but if I need or anyone else need it I'll probably end up implementing it at some point! I mentioned #300 because I needed to add a function here so that I wasn't using a `racket-mode` private function on lispy. so, hum, I don't know the answers to these questions yet! 

> Also, please test the scenario where racket-repl-mode doesn't exist or doesn't have a "live" REPL prompt (no Racket back-end process yet)? Is there a reasonable end user error message?

it just works well, actually! it asks to run the buffer, if user says yes, all works well, if user says no, it says you need to run the REPL first.

greghendershott:
>> Also, please test the scenario where racket-repl-mode doesn't exist or doesn't have a "live" REPL prompt (no Racket back-end process yet)? Is there a reasonable end user error message?

> it just works well, actually! it asks to run the buffer, if user says yes, all works well, if user says no, it says you need to run the REPL first.

Great -- thanks!

greghendershott:
> I have only added support for visiting definition and making RET work on the REPL. I haven't done anything about the lispy-eval capability, mostly because I don't use it very much, specially when using racket. but if I need or anyone else need it I'll probably end up implementing it at some point! I mentioned #300 because I needed to add a function here so that I wasn't using a racket-mode private function on lispy. so, hum, I don't know the answers to these questions yet!

OK I see. That makes sense.

Also that limited thing is fine with me. Honestly, I'm a little "anxious" about the idea of racket-mode attempting to support lispy very broadly. From browsing the issues and code, I get the impression -- an initial impression, admittedly perhaps unfair -- that it is very ambitious, very opinionated, and very DWIM. Which is all great for a user, but it makes me a little nervous about the initial and ongoing effort to interoperate with it broadly and correctly.

TL;DR one baby step at a time, is fine with me.

odanoburu:
> From browsing the issues and code, I get the impression -- an initial impression, admittedly perhaps unfair -- that it is very ambitious, very opinionated, and very DWIM. 

although I've been using it for a while, I only use a very limited subset of it, so I wouldn't say my opinion is much more than an impression either, but I tend to agree with your assessment! I promise not to propose radical changes :crossed_fingers:

racket-edit.el
```diff
@@ -200,14 +200,22 @@ Please keep in mind the following limitations:
...
-                  (not (equal (racket--repl-file-name+md5)
-                              (cons (racket--buffer-file-name t) (md5 (current-buffer)))))
-                  (y-or-n-p "Run current buffer first? "))
-             (racket--repl-run nil nil
-                               (lambda (_n/a)
-                                 (racket--do-visit-def-or-mod 'def str)))
-           (racket--do-visit-def-or-mod 'def str)))))
+    (str (racket--visit-symbol-definition str))))
+
+(defsubst racket-lispy-visit-symbol-definition (str)
```
greghendershott: Should this be a plain old `defun` instead? Genuine question; I'm not sure. (Inlining across packages seems like it could be trouble. e.g. Someday we need to change this. We do. User updates racket-mode package. But. That doesn't mean lispy package necessarily gets recompiled then. Ergo confusion.)

greghendershott:
I'm sorry to be slow to follow up. Traveling last week. Manged to get a head cold over the weekend.

I added a comment/question about the use of `defsubst`. Otherwise I'm OK to merge this. Thanks again.

greghendershott:
Thanks!

odanoburu:
> I added a comment/question about the use of defsubst

you were absolutely right about this, btw! I shouldn't inline things lightly, specially in a case like this.. thanks for the heads-up!

greghendershott:
The funny thing is, I didn't know for sure it was wrong. I just had a hunch.

I recall some weirdness with racket-mode using Emacs Lisp macros -- even just used by/within itself -- when it comes time for users to do package updates. Even just updating just racket-mode, alone. That's what led to me adding a section to the docs, about uninstalling racket-mode, restarting Emacs, and innstalling racket-mode again, to be sure.

I suppose Emacs Lisp is very much from the neighborhood that motivated Racket's "separate compilation guarantee". :smile:

odanoburu:
that and racket's CREPL (`c` is for cleaning State up)! 

I guess reloading everything all the time would make for an unpleasant editing experience, though..

odanoburu:
just for the record, abo-abo acepted https://github.com/abo-abo/lispy/pull/488  and even added (incomplete but still useful) eval functionality :clinking_glasses:

PR #375, opened 2019-05-16
Make sending code as string the primitive REPL interaction (instead of as a region)

countvajhula:
**Summary of Changes**
Currently, the primitive utility function that sends code to the Racket REPL for evaluation is one that extracts a region of interest and sends it via `comint-send-region`. In some cases though, it may be desirable to send code to the Racket REPL that isn't present in the buffer directly -- for instance, a modified version of code that is present in the buffer. In this case, we may need to send a string rather than a region.

Since string is a more elementary representation than a region, it may be preferable to use a string-based utility as the primitive. This PR adds this utility (which is mostly based on the existing region-based primitive), and modifies the current region-based functions to use the new string-based one.

**Context / example usecase:**
While working with some racket code I wanted to write a utility that would allow me to evaluate functions in the Racket REPL, and in certain cases "pretty print" the result if it is not of a type that renders usefully. For this purpose I needed to modify the actual code being evaluated with a wrapped conditional expression like `((let ([result ,<extracted-expression-from-the-buffer>]) (cond [(stream? result) (stream->list result)] ...))` and send this new string to the REPL. To avoid code duplication it seemed that sending a string to the REPL would be better provided by racket-mode than as a custom utility.

### DIFF
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -549,8 +549,7 @@ Afterwards call `racket--repl-show-and-move-to-end'."
 (defun racket--send-region-to-repl (start end)
   "Internal function to send the region to the Racket REPL."
   (when (and start end)
-    (racket--send-to-repl (buffer-substring (region-beginning)
-                                            (region-end)))))
+    (racket--send-to-repl (buffer-substring start end))))
 
 (defun racket-send-region (start end)
   "Send the current region (if any) to the Racket REPL."
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -525,27 +525,32 @@ most recent `racket-mode' buffer, if any."
 
 ;;; send to REPL
 
-(defun racket--send-region-to-repl (start end)
-  "Internal function to send the region to the Racket REPL.
+(defun racket--send-to-repl (code)
+  "Internal function to send CODE to the Racket REPL for evaluation.
 
-Before sending the region, calls `racket-repl' and
+Before sending the code (in string form), calls `racket-repl' and
 `racket--repl-forget-errors'. Also inserts a ?\n at the process
 mark so that output goes on a fresh line, not on the same line as
 the prompt.
 
 Afterwards call `racket--repl-show-and-move-to-end'."
+  (racket-repl t)
+  (racket--repl-forget-errors)
+  (let ((proc (get-buffer-process racket--repl-buffer-name)))
+    (with-racket-repl-buffer
+      (save-excursion
+        (goto-char (process-mark proc))
+        (insert ?\n)
+        (set-marker (process-mark proc) (point))))
+    (comint-send-string proc code)
+    (comint-send-string proc "\n"))
+  (racket--repl-show-and-move-to-end))
+
+(defun racket--send-region-to-repl (start end)
+  "Internal function to send the region to the Racket REPL."
   (when (and start end)
-    (racket-repl t)
-    (racket--repl-forget-errors)
-    (let ((proc (get-buffer-process racket--repl-buffer-name)))
-      (with-racket-repl-buffer
-        (save-excursion
-          (goto-char (process-mark proc))
-          (insert ?\n)
-          (set-marker (process-mark proc) (point))))
-      (comint-send-region proc start end)
-      (comint-send-string proc "\n"))
-    (racket--repl-show-and-move-to-end)))
+    (racket--send-to-repl (buffer-substring (region-beginning)
+                                            (region-end)))))
 
 (defun racket-send-region (start end)
   "Send the current region (if any) to the Racket REPL."


countvajhula:
I'm actually not sure why the tests are failing since the tests passed on the last commit... in Travis it looks like it might be a MELPA issue, but when I run the tests locally with `make test` it looks like it's attempting to connect to a running racket process and getting errors like `(user-error "Current buffer is not a racket-mode buffer")`. It may also be using the version of racket-mode in my `.emacs.d` rather than the cloned repo for this PR since I see messages like `Starting racket to run /Users/siddhartha/.emacs.d/elpa/racket-mode-20181030.1345/racket/run.rkt ...
Connected to racket-command process on port 55557 after 6 attempt(s)`. If you've seen this before and know what's going on, please let me know, otherwise I'll take another crack at investigating soon.

greghendershott:
Just a quick comment until I have time to look at it more carefully:

What you're describing sounds like the Racket `current-print` handler. Could you simply use that?

Let's say you define `my-print-handler`.  You might think, well, wrapping every REPL expression in a `parameterize` is nearly as tedious as wrapping it in a `let`.  But remember, you can enter `(current-print my-print-handler)` once in the REPL -- then subsequent things will use that.

(This will get reset on each `racket-run`. But you could stick the `(current-print my-print-handler)` in the `.rkt` file that you're running.)

greghendershott:
> I'm actually not sure why the tests are failing since the tests passed on the last commit... in Travis it looks like it might be a MELPA issue

I restarted that job. It passed this time.

> It may also be using the version of racket-mode in my .emacs.d rather than the cloned repo for this PR since I see messages like Starting racket to run /Users/siddhartha/.emacs.d/elpa/racket-mode-20181030.1345/racket/run.rkt ... Connected to racket-command process on port 55557 after 6 attempt(s)

Yes. 2018-10-30 is somewhat old compared to `master`. For example, I could also guess that from the "connected after 6 attempts" output.

When developing locally, it's probably best to package-delete then add the git clone path to your Emacs `load-path`. And restart Emacs. That way you'll be sure to use the code from git. And things like `make test` will test that code.

countvajhula:
Got it. I think the ability to use `current-print` may be a separate issue from this PR. The way I'm thinking about it is that with "sending code to the REPL for evaluation" as the overall operation that is happening here, the most natural representation of the code being sent is as text, rather than as a buffer region, since all regions have corresponding text but not all text is present in a buffer. This would give more flexibility in sending code to the Racket REPL since the existing region-based behavior is trivially modeled as text, and we could also support other usecases like the one I mentioned earlier, or conceivably others like:

- obtaining Racket code from another process and sending it to the REPL directly
- making modifications to code from a buffer in sending it to the REPL that are not printing-related. E.g. we may want to prepend all variable names with `test-` for some reason, in sending code to the REPL
- we might also want to have some emacs keybindings to execute some specific racket code when invoked, e.g. code to display the current active bindings or parameters, or something like this which we may want to do periodically as a status check that has nothing to do with the actual code we're writing/testing

countvajhula:
Re: `current-print`, if I'm understanding it correctly, I can't do it one-time in my particular case since only the "pretty print" command is meant to use this custom printing behavior, and normally it should print the usual way unless this functionality is specifically invoked. So, if I'm wrapping each expression on demand, I think I would still need to send it to the REPL as text rather than as a region?

greghendershott:
A few things.

Be aware that `sequence?` means, "Can this be _used as_ a sequence?". So e.g. `(sequence? 10)` is true. So your `cond` that does `sequence->list` will print `(list 1 2 3 4 5 6 7 8 9 10)`.

You could add a `number?` test first. I'm not sure that exhausts the list of usable-as-a-sequence things. But let's assume it is, for the following.

---

I think your print handler could be something like this:

```racket
(require racket/sequence
         racket/stream)

(define (make-listy-print-handler)
  (define old-handler (current-print))
  (define (new-handler v)
    (old-handler
     (cond [(number? v)   v]  ;avoid (sequence? number) true
           [(sequence? v) (sequence->list v)]
           [(stream? v)   (stream->list v)]
           [else          v])))
  new-handler)
```
Note that this remembers and delegates to the original print handler, which can still do whatever special or boring things that it wants to do.

Now you can use it in your Racket file or in the REPL like this:

```racket
(current-print (make-listy-print-handler))
```

For example in the REPL:

```
example.rkt> (in-range 10)
#<stream>
example.rkt> (current-print (make-listy-print-handler))
example.rkt> (in-range 10)
'(0 1 2 3 4 5 6 7 8 9)   ;now the stream prints as a list
```
---

In summary, I think this is best handled in your Racket code, not in racket-mode. As a result, although I appreciate you taking the time to make and discuss this PR, I'd prefer not to merge it.

countvajhula:
Right, I only provided the sequence pretty printing as an example -- that particular feature, i.e. the pretty printing feature, is very specific to my workflow and I don't expect it to be widely useful in any way. I only meant that in sending code to the racket REPL I'd prefer to be able to send Racket mode a string and have it handle sending it to the REPL, instead of writing my own string to REPL converter, which currently looks like this in my `.emacs.d`:

```
(defun racket--send-to-repl (code)
  "(Keeping this here for now pending possible incorporation into racket-mode)
Internal function to send CODE to the Racket REPL for evaluation.

Before sending the code (in string form), calls `racket-repl' and
`racket--repl-forget-errors'. Also inserts a ?\n at the process
mark so that output goes on a fresh line, not on the same line as
the prompt.

Afterwards call `racket--repl-show-and-move-to-end'."
  (racket-repl t)
  (racket--repl-forget-errors)
  (let ((proc (get-buffer-process racket--repl-buffer-name)))
    (with-racket-repl-buffer
      (save-excursion
        (goto-char (process-mark proc))
        (insert ?\n)
        (set-marker (process-mark proc) (point))))
    (comint-send-string proc code)
    (comint-send-string proc "\n"))
  (racket--repl-show-and-move-to-end))
```
Notice it's almost verbatim the region-to-repl code in Racket mode, hence my suggestion to include such functionality in racket mode and this PR, since most of this is not specific to my workflow but rather to the details of executing _generic_ Racket code provided as a string.

I appreciate the pointer to `current-print` -- I didn't know about it before and it looks useful, and your example code to retain the old handler is neat! But it doesn't do what I'd like since it is only _on demand_ that I'd like to pretty print certain sequences and not all sequences for some of the reasons you mentioned (like `5` being a sequence, but also e.g. some sequences being infinite streams). Really my pretty-print feature is not at all what this PR is about. It's just about sending text to the REPL that represents generic racket code, and whether that should be something Racket mode should provide, or whether that's something anyone requiring it would have to implement like I did above. I personally feel the former is warranted since otherwise there is duplication of what is already present in racket mode, serving as it does as the liaison between the user and the REPL. But if you feel otherwise -- and I gladly leave it to your better judgement here -- then please feel free to close this PR 😺

countvajhula:
I'm closing this PR for now. Thanks for taking the time to discuss it.

PR #387, opened 2019-07-07
Remove wrong keymapping.

wouteroostervld:
Reverts the keymapping "[C-S-backspace]" fixing the inserting of the leading square bracket "[" in the repl.

### DIFF
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -714,7 +714,6 @@ With prefix arg, open the N-th last shown image."
      ("C-M-q"           prog-indent-sexp)
      ("C-a"             comint-bol)
      ("C-w"             comint-kill-region)
-     ("[C-S-backspace]" comint-kill-whole-line)
      (")"               racket-insert-closing)
      ("]"               racket-insert-closing)
      ("}"               racket-insert-closing)


wouteroostervld:
The removed keymapping is not functioning anyway (tested on spacemacs). Maybe a commit should follow with a fix which implements the original intention.

wouteroostervld:
Travis has trouble fetching gnutls (at this moment).

greghendershott:
Thank you for the PR.

I hadn't noticed the problem in daily use because it doesn't occur with paredit-mode, which I use.

I think the root issue is more simply that the wrong keybinding notation is used in `racket-repl-mode-map`, and a better fix would be this:

```diff
modified   racket-repl.el
@@ -714,7 +714,7 @@ With prefix arg, open the N-th last shown image."
      ("C-M-q"           prog-indent-sexp)
      ("C-a"             comint-bol)
      ("C-w"             comint-kill-region)
-     ("[C-S-backspace]" comint-kill-whole-line)
+     ("<C-S-backspace>" comint-kill-whole-line)
      (")"               racket-insert-closing)
      ("]"               racket-insert-closing)
      ("}"               racket-insert-closing)
```

Could you please confirm this works for you, too?

If so: Please either revise the PR to do that, or, let me know if you'd like me to make the change. (Either way, thank you again for your time to at least prepare this original PR!)

If not: Let's discuss more.

greghendershott:
As more background: The bad binding used to be masked by an explicit binding for <kbd>[</kbd> to `racket-smart-open-bracket`. However that was removed a couple weeks ago in commit 669275b -- letting the bad binding now cause a problem.

Having <kbd>[</kbd> not work properly in the REPL (unless you're using paredit) is a pretty basic, bad, embarrassing bug on my part. Therefore instead of waiting for you to revise this PR, I think I will go ahead and push a commit with the fix I describe above -- and thank you in the commit message!

greghendershott:
Thanks again! Of course please let me know if my commit doesn't resolve things for you, too.

wouteroostervld:
Yes, this fixes it. Thanks!

PR #421, opened 2020-02-06
check-syntax: fix arity for def/dr-jump

Bogdanp:
This may not be the right fix here so feel free to close it. I just noticed the problem and threaded the missing argument through.


greghendershott:
Thanks! Instead of merging I'll probably tag this to be closed when I push a commit making a similar change. (Great minds think alike. I'd noticed I recently broke this when making some other changes to keep "describe" working for the "old" racket-run module->namespace scenario, too.)

PR #422, opened 2020-02-10
Add eldoc support via scribble/blueboxes

xuchunyang:
I noticed DrRacket's hover feature is built on https://docs.racket-lang.org/scribble/blueboxes.html, I think we can also try it for the eldoc support.

### DIFF
--- a/racket-complete.el
+++ b/racket-complete.el
@@ -193,6 +193,29 @@ A more satisfying experience is to use `racket-describe' or
                str)
            (scan-error nil)))))
 
+(defun racket-blueboxes-eldoc-function ()
+  (and (racket--repl-live-p)
+       (save-excursion
+         (condition-case nil
+             ;; The char-before and looking-at checks below are to
+             ;; avoid calling `racket--get-type' when the sexp is
+             ;; quoted or when its first elem couldn't be a Racket
+             ;; function name.
+             (let* ((beg (progn
+                           (backward-up-list)
+                           (and (not (memq (char-before) '(?` ?' ?,)))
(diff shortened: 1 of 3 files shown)


greghendershott:
Interesting. I was under the impression that people expect eldoc to show just one line in the echo area. Am I mistaken?

xuchunyang:
No, you're right, the echo area usually is one line height, so it's a good practice to not passing multiline text to `message`, however, sometime it's impossible.

greghendershott:
Well, I guess there is no harm in providing this as an option for people to try using as an eldoc function.

This pull request is awkward to merge right now, because my `check-syntax` branch has diverged in some key respects. Rebasing on this might be harder than me just using your PR as a guide to doing something similar on the new branch. So if you don't mind, I think I'll do that, and tag this PR in a commit message to be closed, once the check-syntax branch is merged to `master`.

xuchunyang:
> So if you don't mind, I think I'll do that


I don't mind. 

I didn't expect the PR to be merged, e.g., the code of `racket-blueboxes-eldoc-function` is copy-and-pasted from `racket-eldoc-function`, there is duplicated code, and it doesn't work all the time, for example

```racket
;; unexpected: show nothing
(version)

;; expected: show the sig
(version )
        ^ a space is required for some reason
```

greghendershott:
Great. In that case I'll leave the PR open, until something is merged to `master`,  because it's a good enhancement idea.

greghendershott:
This is closed via commit 47b9041. Thanks again for providing the idea and head-start.

PR #427, opened 2020-02-20
Use open as racket-images-system-viewer for macOS

xuchunyang:
I notice `M-x racket-view-last-image` does not work for me until I change the value of `racket-images-system-viewer` to the open(1) command.


greghendershott:
Thank you!

Issue #432, opened 2020-03-01
racket--font-lock-sexp-comments does not match racket behavior


tgbugs:
Paste the following block into a racket-mode buffer.
```racket
#; ; this comments out the second sexp
   ; the racket reader skips over the intervening comment
   ; because it is a comment (I think)
#; ; this comments out the first sepx
(first sexp to be commented out)
(second sexp to be commented out is not highlighted)
```
If you run the code you will not get the error you would expect based on fontification.
The second sexp is not highlighted despite the fact that racket does not execute it.
I encountered this while writing tests to try and replicate racket-mode's fontification for a common lisp reader macro that does the same thing.

This is a fontification edge case.
The smallest repro is
```racket
#; #; 'one 'two 'three
```

<details>
```
((alist-get 'racket-mode package-alist)
 #s(package-desc racket-mode
                 (20200115 2148)
                 "Major mode for Racket language."
                 ((emacs
                   (24 3))
                  (faceup
                   (0 0 2)))
                 nil nil "/home/tom/.emacs.d/elpa/racket-mode-20200115.2148"
                 ((:url . "https://www.racket-mode.com/")
                  (:maintainer "Greg Hendershott")
                  (:authors
                   ("Greg Hendershott")))
                 nil)
 #s(package-desc racket-mode
                 (20191204 205)
                 "Major mode for Racket language."
                 ((emacs
                   (24 3))
                  (faceup
                   (0 0 2)))
                 nil nil "/home/tom/.emacs.d/elpa/racket-mode-20191204.205"
                 ((:url . "https://www.racket-mode.com/")
                  (:maintainer "Greg Hendershott")
                  (:authors
                   ("Greg Hendershott")))
                 nil))
((emacs-version "26.3")
 (emacs-uptime "19 hours, 20 minutes, 38 seconds")
 (system-type gnu/linux)
 (major-mode racket-mode)
 (racket--el-source-dir "/home/tom/.emacs.d/elpa/racket-mode-20200115.2148/")
 (racket--rkt-source-dir "/home/tom/.emacs.d/elpa/racket-mode-20200115.2148/racket/")
 (racket-program "racket")
 (racket-command-port 55555)
 (racket-command-startup nil)
 (racket-command-timeout 10)
 (racket-memory-limit 2048)
 (racket-error-context medium)
 (racket-retry-as-skeleton t)
 (racket-error-context medium)
[...]

PR #433, opened 2020-03-02
sexp comment fontification

tgbugs:
This PR should resolve the sexp comment fontification issues noted in #432, however it may introduce additional issues. I have not tested this yet beyond checking comments.rkt to see if the fontification is correct, but wanted to get the implementation out there for review and comparison given the discussion thread in #432. Once `racket-forward-sexp` is implemented it is ok to continue to mark sexp comments as quotes syntactically since the behavior is more in line with what users would expect when navigating.

### DIFF
--- a/racket-font-lock.el
+++ b/racket-font-lock.el
@@ -241,6 +241,38 @@ doesn't really fit that.")
               racket-here-string-face)))
       font-lock-comment-face)))
 
+;;; sexp comment helper machinery from `slime'
+
+(defmacro racket-point-moves-p (&rest body)
+  "Execute BODY and return true if the current buffer's point moved."
+  (declare (indent 0))
+  (let ((pointvar (cl-gensym "point-")))
+    `(let ((,pointvar (point)))
+       (save-current-buffer ,@body)
+       (/= ,pointvar (point)))))
+
+(defun racket-forward-sexp (&optional count)
+  "Like `forward-sexp', but understands sexp comments (#;),
+and skips comments."
+  (dotimes (_i (or count 1))
+    (racket-forward-cruft)
+    (forward-sexp)))
+
+(defun racket-forward-cruft ()
+  "Move forward over whitespace, comments, reader conditionals."
+  (while (racket-point-moves-p (skip-chars-forward " \t\n")
+                               (forward-comment (buffer-size))
+                               (inline (racket-forward-sexp-comment)))))
+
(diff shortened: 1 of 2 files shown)


greghendershott:
Thank you! I'm sorry to be slow to follow up on this. I was distracted by wrapping up the "check-syntax" branch, and then also... world events.

I fetched this locally to take a quick look.

Good news, this rebases onto `master` (with 200 commits from `check-syntax` merged) cleanly.

For testing Emacs font-lock, I've been [putting example files in `racket/example` and using `faceup`](https://github.com/greghendershott/racket-mode/blob/master/CONTRIBUTING.md#font-lock). I think I'd like to use that approach, for this. Would you be interested in doing that, following the existing examples?

If you can, that would be great because you already have a clear understanding/expectation of what the correct result should look like. Then you can "snapshot" that using M-x faceup. Whereas I think I'd be more likely to make a mistake. However if you don't have bandwidth to do this, I'll try.

greghendershott:
p.s. Instead of adding new `example/comment.rkt` and `example/comment.rkt.faceup` files: You could add more s-expression comment examples to the existing `example/example.rkt`, regenerate its `example/example.rkt.faceup` file, and commit those. That might be less "busy work" for you. Either way is fine, for me.

tgbugs:
I will try to take a look at this by sometime this coming weekend. Moving the contents of comment.rkt -> example.rkt is probably the approach I will take, I added comment.rkt separately because I didn't know if adding to examples might cause an issue. If it won't then simple enough to put everything in one place.

greghendershott:
This morning I decided I could help by adding the examples to `examples.rkt` and `examples.rkt.faceup`, myself, and asking you just to eyeball / double-check it.

I did that, then as I was looking more at the code to think about additional test cases, I realized I could approach it in a way that is (for me) somewhat simpler. I just pushed that as commit ad5f9dd on a topic branch. If/when I merge that to `master`, it will close this PR as well as issue #432.

greghendershott:
As the commit message says, big thanks for filing the issue, showing some examples and test cases, and proposing a solution!!

tgbugs:
Everything looks good on my end. You have a much better (the best possible?) idea about where and how it is most efficient to implement something like this. One advantage of your approach is that implementation for font locking stays decoupled from the implementation for how things are sent to the repl. Glad I was able to get the ball rolling. Thanks!

PR #434, opened 2020-03-24
Add asterisks to buffer names that don't correspond to files

sorawee:
Following the convention described at https://www.emacswiki.org/emacs/Buffer

### DIFF
--- a/racket-cmd.el
+++ b/racket-cmd.el
@@ -72,8 +72,8 @@ See issue #327.")
    :name            racket--cmd-name
    :connection-type 'pipe
    :noquery         t
-   :buffer          (get-buffer-create racket--cmd-name)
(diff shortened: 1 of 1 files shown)


greghendershott:
I tend to do this with user buffers, e.g. `*Racket Logger*` or whatever.

I didn't do this for the process buffers because I've seen counter-examples, e.g. `magit-process`. 

Also, when using `M-x list-processes`, which truncates process names in its display, you can distinguish `racket-process` from `racket-process-stderr`, whereas putting a `*` on the front causes the truncation to make them indistinguishable -- which was a big deal to me a month ago when debugging the new back end process approach. :smile:

Having said all that, I guess the asterisks would be better. I might wait to merge this until I sort through a few other things, just in case it turns out to be a non-trivial change.

PR #453, opened 2020-04-23
Recognize srcloc structs in REPL

vkz:
We extend compilation-error-regexp-alist with a recognizer for srcloc
structs that get linkified and highlighted as INFO. This won't
distract you from normal work but will improve quality of life when
you work with parsers or readers that track source locations the
"Racket way". This tiny patch lets you immediately jump to
corresponding location.


greghendershott:
Thanks!

Racket installs on Travis CI have been flaky for a few days. That's why your PR didn't pass checks. Also that's why I didn't reply yet -- trying to sort out why.

Will definitely merge this.

Issue #454, opened 2020-04-25
Hacking on Racket Mode: what is the typical development setup?


scolobb:
Hello everyone,

I would like to add a function `racket-xp-helpdesk` which would correspond to "Search in Help Desk" menu item in DrRacket [0].  I've been trying to point my Emacs to the clone of this repository for some time now, and I'm trying heavier and heavier solutions (from my subjective perspective).  I'll eventually figure out how to do it, but I wonder what is your typical setup for hacking on Racket Mode.

Do you add the directory of the Git repository to `load-path` and then do a bunch of requires?  Is it actually enough?  Is there an easier way to import the whole of Racket Mode without uninstalling the version coming from my package manager (Nix)?

------

[0] I've seen that `racket-xp-documentation` actually falls back to calling the `search` function from the Racket backend when it cannot find the documentation of a particular symbol, but I'd like to be able to explicitly open the search page.  My use case is as follows.  When I use Typed Racket, `racket-xp-documentation` will show me the documentation pages from Typed Racket, but I often want too see the documentation from `racket` because it gives more details.

greghendershott:
It should be as simple as adding the git repo directory to your `load-path`.

For example if you use `use-package`, you can simply replace

    (use-package racket-mode
       :ensure t)

with

    (use-package racket-mode
      :load-path "/path/to/the/git-clone/dir")

Now, you might _also_ want to:

- `M-x package-delete` `racket-mode` so the stuff under `.emacs.d/elpa` doesn't get loaded.
- Restart Emacs, to be sure.

scolobb:
Thank you @greghendershott ! I had trouble doing `(require 'racket-mode)` even when I added the Git checkout to my `load-path`, and I have just realized that I didn't have `pos-tip` installed :facepalm: I installed it and everything works fine now.

I really should have read `CONTRIBUTING.md` more attentively.

PR #455, opened 2020-04-27
CONTRIBUTING.md: Explain how to point Emacs to the Git checkout.

scolobb:
This PR extends `CONTRIBUTING.md` with some additional tips about setting up Emacs for hacking on Racket Mode.

It summarizes my own experience as well as @greghendershott 's advice from Issue https://github.com/greghendershott/racket-mode/issues/454.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,6 +35,43 @@ sure your PR updates all of:
 1. the `Package-Requires:` line in `racket-mode.el`
 2. the `deps` target in `makefile`
 
+## Pointing Emacs to your Git clone
+
+After ensuring all dependencies of Racket More are installed, it
+suffices to add the path to your local clone of Racket Mode to
+`load-path` and require the package:
+
+```elisp
+(add-to-list 'load-path "/path/to/the/git-clone/dir")
+(require 'racket-mode)
+```
+
+Note that these lines will override any previous references to Racket
+Mode in your Emacs configuration.  In particular, if you have Racket
+Mode installed as an Emacs package, after evaluating these lines you
+will use Racket Mode from your local Git repository.
+
+If you use `use-package`, you can simply replace
+
+```elisp
+(use-package racket-mode
+    :ensure t)
+```
+
+with
+
+```elisp
+(use-package racket-mode
+    :load-path "/path/to/the/git-clone/dir")
+```
+
+You might also need to:
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -35,6 +35,45 @@ sure your PR updates all of:
 1. the `Package-Requires:` line in `racket-mode.el`
 2. the `deps` target in `makefile`
 
+## Pointing Emacs to your Git checkout
```
greghendershott, line 38: Instead of "checkout" should this be "clone"?

CONTRIBUTING.md
```diff
@@ -35,6 +35,45 @@ sure your PR updates all of:
...
+```
+
+with
+
+```elisp
+(use-package racket-mode
+    :load-path "/path/to/the/git-clone/dir")
+```
+
+You might also need to:
```
greghendershott, line 70: This advice is supposed to apply to both methods -- using `(add-to-path 'load-path _)` or `(use-packge :load-path __)`. But with the big "Using use-package" heading, it feels like you're saying it only applies to that method?  Maybe just delete that heading?

CONTRIBUTING.md
```diff
@@ -35,6 +35,45 @@ sure your PR updates all of:
...
+(add-to-list 'load-path "/path/to/the/git-clone/dir")
+(require 'racket-mode)
+```
+
+Note that these lines will override any previous references to Racket
+Mode in your Emacs configuration.  In particular, if you have Racket
+Mode installed as an Emacs package, after evaluating these lines you
+will use Racket Mode from your Git checkout.
+
+### Using `use-package`
```
greghendershott, line 54: Maybe delete this heading? See other comment below.

greghendershott:
Thanks! I made a few nit-picky suggestions. Eye-of-the-beholder stuff; writing docs is hard!

CONTRIBUTING.md
greghendershott: By "this advice" I meant lines 70-76 -- which is advice I'd originally given you. I'm not even sure if it's strictly necessary. Sometimes Emacs can feel like a big pile of mutable state, which I why I advise deleting the MELPA package and restarting Emacs, just to be sure.

Anyway, the advice was meant for both ways of changing the load-path -- either directly or via use-package.

greghendershott:
Thanks -- merged!

PR #458, opened 2020-04-30
Fix typo in racket-mode.org

pmatos:
Fixed analysis

PR #459, opened 2020-04-30
Fix typo in racket-mode.texi

pmatos:
Fixed analysis.

Similar to #458


greghendershott:
Thanks! Closed by commit 972ba13.

PR #462, opened 2020-05-03
Add `racket-xp-helpdesk` and `racket-repl-helpdesk`

scolobb:
This PR adds two functions to search identifiers (or strings) in the Search Manuals page.  This corresponds to the "Search in the Help Desk" functionality in DrRacket.  These functions mostly exploit the preexisting function `search` in `racket/commands/help.rkt`.

These PR also adds two customizable variables to control the behavior of these functions.

### DIFF
--- a/doc/generate.el
+++ b/doc/generate.el
@@ -62,6 +62,7 @@
     racket-xp-visit-definition
     racket-xp-describe
     racket-xp-documentation
+    racket-xp-helpdesk
     "Run"
     racket-run
     racket-run-and-switch-to-repl
@@ -69,6 +70,7 @@
     racket-repl
     racket-repl-describe
     racket-repl-documentation
+    racket-repl-helpdesk
     racket-repl-visit-definition
     racket-racket
     racket-profile
@@ -167,6 +169,7 @@
     racket-browse-url-function
     racket-xp-after-change-refresh-delay
     racket-xp-highlight-unused-regexp
+    racket-documentation-search
     "REPL variables"
     racket-repl-buffer-name-function
     racket-history-filter-regexp
--- a/racket-custom.el
+++ b/racket-custom.el
@@ -161,6 +161,23 @@ an underline, which is a common convention."
   :safe #'stringp
   :group 'racket-xp)
 
+(defcustom racket-documentation-search
+  "https://docs.racket-lang.org/search/index.html?q=%s"
+  "Where `racket-xp-helpdesk' and `racket-repl-helpdesk' should
+look for the search page.  If the value of this variable is
+'local, open the search page from the local documentation.
+Otherwise, use the URL given by this variable.
+
+This variable should contain a string recognizable by `format',
+with %s at the point at which to insert the identifier to look up
+in the help desk.  Apart from %s, the string should be a properly
+encoded URL."
+  :tag "Where to look for the help desk"
+  :type '(choice (string :tag "URL")
+                 (const :tag "Local" 'local))
+  :safe (lambda (val) (or (stringp val) (eq val 'local)))
+  :group 'racket-xp)
+
 ;;; REPL
 
 (defgroup racket-repl nil
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -936,6 +936,7 @@ by the file module -- not locals or definitions in submodules."
      ("C-c C-e r"       racket-expand-region)
      ("M-C-y"           racket-insert-lambda)
      ("C-c C-d"         racket-repl-documentation)
+     ("C-c C-s"         racket-repl-helpdesk)
      ("C-c C-."         racket-repl-describe)
      ("M-."             racket-repl-visit-definition)
      ("C-M-."           racket-visit-module)
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -898,6 +898,26 @@ instead of looking at point."
                                                  `(doc namespace ,str)
                                                  #'racket-browse-url))))
 
+(defun racket-repl-helpdesk (&optional prefix)
+  "Show the Search Manuals page with the identifier at point.
+
+If `racket-documentation-search' is set to 'local, opens the
+Search Manuals page from the local documentation.  Otherwise,
+opens the page given by the URL in `racket-documentation-search'.
+
+Like `racket-repl-documentation', given a prefix, prompts for the
+identifier, in which case it assumes definitions in or imported
+by the file module -- not locals or definitions in submodules."
+  (interactive "P")
+  (pcase (racket--symbol-at-point-or-prompt prefix "Documentation for: "
+					    racket--repl-namespace-symbols)
+    ((and (pred stringp) symb)
+     (pcase racket-documentation-search
+       ((and (pred stringp) url)
+        (browse-url (format url symb)))
+       ('local
+        (call-process racket-program nil 0 nil "-l" "raco" "doc" symb))))))
+
 ;;; racket-repl-mode
 
 (defvar racket-repl-mode-map
--- a/racket-xp.el
+++ b/racket-xp.el
@@ -58,7 +58,8 @@ everything. If you find that too \"noisy\", set this to nil.")
    `(("C-c #"     ,racket-xp-control-c-hash-keymap)
      ("M-."       ,#'racket-xp-visit-definition)
      ("C-c C-."   ,#'racket-xp-describe)
-     ("C-c C-d"   ,#'racket-xp-documentation))))
+     ("C-c C-d"   ,#'racket-xp-documentation)
+     ("C-c C-s"   ,#'racket-xp-helpdesk))))
 
 (easy-menu-define racket-xp-mode-menu racket-xp-mode-map
   "Menu for `racket-xp-mode'."
--- a/racket-xp.el
+++ b/racket-xp.el
@@ -480,6 +480,26 @@ definitions in submodules."
                            `(doc ,(buffer-file-name) ,str)
                            #'racket-browse-url))))))
 
+(defun racket-xp-helpdesk (&optional prefix)
+  "Show the Search Manuals page with the identifier at point.
+
+If `racket-documentation-search' is set to 'local, opens the
+Search Manuals page from the local documentation.  Otherwise,
+opens the page given by the URL in `racket-documentation-search'.
+
+Like `racket-xp-documentation', given a prefix, prompts for the
+identifier, in which case it assumes definitions in or imported
+by the file module -- not locals or definitions in submodules."
+  (interactive "P")
+  (pcase (racket--symbol-at-point-or-prompt prefix "Documentation for: "
+					    racket--xp-binding-completions)
+    ((and (pred stringp) symb)
+     (pcase racket-documentation-search
+       ((and (pred stringp) url)
+        (browse-url (format url symb)))
+       ('local
+        (call-process racket-program nil 0 nil "-l" "raco" "doc" symb))))))
+
 (defun racket-xp--forward-use (amt)
   "When point is on a use, go AMT uses forward. AMT may be negative.
 


greghendershott:
Thank you for submitting this! I haven't had a chance to take a look, yet -- so I wanted to let you know I do have it on my to-do list and will try to get to it soon!

scolobb:
Sure, no problem, thank you for telling me.  Take your time!

greghendershott:
I'm sorry to be so slow to follow up on this.

I want to make sure I understand the motivation.

The status quo:

With the existing help commands bound to <kbd>C-c C-d</kbd> -- `racket-xp-documentation` in `racket-xp-mode`, as well as `racket-repl-documentation` in `racket-repl-mode` -- the status quo idea is:

- <kbd>C-c C-d</kbd> assume the symbol at point, try to find doc for it
- <kbd>C-u C-c C-d</kbd> prompt for the symbol, try to find doc for it

And then in both cases:

- If doc is found, it's shown.
- If doc isn't found -- or if it's ambiguous, e.g. various libs have doc for their thing named "foo" -- you're taken to the Search page with the symbol filled in for you.

---

That is how it's _supposed_ to work. And if things work that way, I'm not sure I understand the need for these new commands? I _think_ this is covered by <kbd>C-u C-c C-d</kbd>, already??

So:

- Maybe you encountered a bug, which led you to think we need these new commands... but maybe instead I should fix this bug? :smile:  (It's possible I broke something when adding `racket-xp-mode`, but haven't noticed in my own use, yet.)

- Maybe I'm just totally misunderstanding the motivation, and you can help me understand better, and we should add the new commands, too. :smile:

greghendershott:
Is the motivation an example like this:

You have some file or REPL that is `#lang racket`. That has a certain `define`. If you ask for doc about that, it's not ambiguous, so you'll get that only/automatically. No annoying stop at the Search page, first, to confirm.

But maybe you're curious about all the `define`s various things provide and document. You _want_ the Search page.

Is this what you have in mind?

scolobb:
> But maybe you're curious about all the `define`s various things provide and document. You _want_ the Search page.

Yes, the principal use case is exactly this.  In fact, my original problem is the following.  Consider a file containing this code:

```racket
#lang typed/racket

(for/list : (Listof Integer) ([i (in-range 10)]) i)
```

If I run `racket-xp-documentation` with the point on `for/list`, I get the documentation page for Typed Racket variants of this syntax, which says that it works in the same way as its untyped equivalent.  Not very helpful if I want to see some actual examples (e.g., because I forgot how this particular macro is used, or in which direction it traverses the list, etc.), so I need to manually look for the untyped variant of `for/list`.  If I use `racket-xp-helpdesk`, I can directly choose the untyped definition among the first search results.

Another variant of this use case (I believe it's the use case you are talking about): I would like to see which libraries define a particular symbol and choose one depending on my needs.

Finally, another use case, which is probably slightly collateral but I do need this from time to time, is a situation in which I want a symbol from a package which is *not* installed on my machine.  `racket-xp-documentation` always opens the local search page, which looks into the docs of *installed* packages.  `racket-xp-helpdesk` on the other hand defaults to opening the online search page which covers more packages, and helps me remember what package to install to get the symbol I need.

scolobb:
I'll be on vacation until August 27 and I'll be able to handle all the changes/comments when I'm back.

greghendershott:
I looked at your PR in more detail. It looks really good. I think my only two suggestions are:

1. The `.texi` has many diffs which are unrelated to your change. I think because you're using a newer version of Emacs. Probably there are harmless but it makes the diff have a low signal:noise ratio.  Could you change that commit only to change `generate.el`? I'll regenerate the texi on my end.

2. I totally understand why you looked in `help.rkt` as a starting point. I would have, too. But I don't think adding a new back end `search` command makes sense. Instead: Either the user wants the online `docs.racket-lang.org` search page, in which case we can just call `browse-url`, -OR-, they want local help, in which case we can just invoke `raco docs <search terms>` and it will open the browser for whatever local doc dir is appropriate.

    Suggestion: Add a single customization variable, `racket-documentation-search`. It can be the value `'local`, in which case we do a `raco doc`. Otherwise it's a format string, like the var you named `racket-helpdesk-url`, used to call `browse-url`. As a default I guess I'd suggest it be `https://docs.racket-lang.org/search/index.html?q=%s`.

    So effectively the same as what you have now, albeit using 1 var instead of 2. And not using the Racket Mode back send.

    Am I explaining that well? What do you think?

scolobb:
Thank you for feedback!  No problem with (1) and I think I understand and agree with (2).  I'll start doing the changes tonight, but it's probably going to take a couple of days (deadlines, start of the school year, etc.).  I'll push the updates and I'll tell you when I'm done.

greghendershott:
Welcome back from vacation! :smile:

If things are too busy for you with deadlines and school, and it turns out you don't have time? Then let me know and I can try to 
make the finishing changes. I feel bad I wasn't able to review this earlier in the summer.

scolobb:
> Welcome back from vacation! :smile:

Haha, thank you! :smile: 

> If things are too busy for you with deadlines and school, and it turns out you don't have time? Then let me know and I can try to
> make the finishing changes.

Thank you!  I expect to have some time to work on this PR in the following days, and anyway I'm always happy to learn something about the inner workings of Racket Mode.

>  I feel bad I wasn't able to review this earlier in the summer.

Nah, don't worry, I don't expect you to be always available for PR review, I can totally understand that!

scolobb:
OK, so I think I implemented both of your suggestions.  In fact (2) hugely simplified the code, nice catch!

I'm still not completely sure of several things in this PR, in particular:

1. Is it OK to write `:type 'sexp` in the definition of `racket-documentation-search`?  Is there a better way to capture the union type `'local OR stringp` in `defcustom`?  I looked around the docs but found nothing of the kind.

2. I run "raco doc` via Emacs' `shell-command`.  Is that right?  This is how I typically run shell commands in my Elisp snippets, but there are maybe better ways (as in "more control over the side effects").

greghendershott:
Thanks -- looking good so far!

> Is it OK to write :type 'sexp in the definition of racket-documentation-search? Is there a better way to capture the union type 'local OR stringp in defcustom? I looked around the docs but found nothing of the kind.

I think `:type 'sexp` is not bad. However you could use `choice`. There are a few examples in `racket-custom.el`. I think the closest example would be this:

```elisp
(defcustom racket-xp-after-change-refresh-delay 1
  "Seconds to wait before refreshing `racket-xp-mode' annotations.

Set to nil to disable automatic refresh and manually use `racket-xp-annotate'."
  :tag "Racket XP Mode After Change Refresh Delay"
  :type '(choice (integer :tag "Seconds")
                 (const :tag "Off" nil))
  :safe #'integerp
  :group 'racket-xp)
```

So maybe something like `(choice (string :tag "URL") (const :tag "Local" 'local))`?  (I just typed that off the top of my head, not tested).

The nice thing about `choice` and the `:tag` things is it influences the TUI for `M-x customize`, which some people prefer.

> I run raco doc via Emacs' shell-command. Is that right? This is how I typically run shell commands in my Elisp snippets, but there are maybe better ways (as in "more control over the side effects").

- I think `shell-command` is the right thing to use (or maybe `start-process`... I'd have to double-check).

- Either way, we have a customization variable `racket-program`, that says how to run `racket`. You should use that. How to run `raco`, though? There's a "trick", `raco <args>` is just shorthand for `racket -l raco <args>`.

- One final point: You probably want to take the user's search term and wrap it in `shell-quote-argument`.

greghendershott:
Actually `shell-command` seems not the right thing:

> shell-command is an interactive and compiled function defined in
> simple.el.gz.
>
> Signature
> (shell-command COMMAND &optional OUTPUT-BUFFER ERROR-BUFFER)
>
> Documentation
> Execute string COMMAND in inferior shell; display output, if any.
>
> With prefix argument, insert the COMMAND's output at point.
>
> Interactively, prompt for COMMAND in the minibuffer.
>
> If COMMAND ends in &, execute it asynchronously.
> The output appears in the buffer *Async Shell Command*.
> That buffer is in shell mode.  You can also use
> async-shell-command that automatically adds &.
>
> Otherwise, COMMAND is executed synchronously.  The output appears in
> the buffer *Shell Command Output*.  If the output is short enough to
> display in the echo area (which is determined by the variables
> resize-mini-windows and max-mini-window-height), it is shown
> there, but it is nonetheless available in buffer *Shell Command
> Output* even though that buffer is not automatically displayed.
>
> ...
>
> In Elisp, you will often be better served by calling call-process or
> start-process directly, since it offers more control and does not impose
> the use of a shell (with its need to quote arguments).

We don't want a buffer to appear showing the non-interesting output from `raco doc`. We're only interested in having `raco doc` do its thing to show the web page in your browser.

Instead I think we want to follow the hint in the last paragraph.

I think this will work:

```elisp
(call-process racket-program nil 0 nil "-l" "raco" "doc" user-search-string)
```

That worked in a quick experiment for me.

My understanding is that `shell-quote-argument` is **not** needed because `call-process` bypasses any shell.

scolobb:
Thank you for you detailed suggestions and explanations!  As you suggested, I fixed the `:type` and `:safe` in the definition of `racket-documentation-search` and modified `racket-xp-helpdesk` and `racket-repl-helpdesk` to run `racket -l raco doc`.

I didn't initially notice you second commend on `call-process`, that's why there are two sets of commits appearing in the history of this PR.  I used to rely on `shell-command` quite a bit, but I now realize that `call-process` may be a better tool in some of my use cases.  Good to know!

greghendershott:
Looks great!

Before I merge: I'm wondering about "refactoring" this, both to share more implementation code, as well as to have fewer commands overall. In other words, the "helpdesk" command is "only" a variation of the main documentation command, that simply means "_always_ search".

I made a branch to flesh this out. It is your commits, rebased onto latest `master`, plus one new commit of my own: https://github.com/greghendershott/racket-mode/commit/849d26efb03f11186636a51bb188c956db590af8

Overall I like it. I'm not sure about using a double <kbd>C-u C-u</kbd> prefix. I like that it doesn't use up another key binding. On the other hand, does that make it too "obscure"? I don't know.

Any thoughts?

greghendershott:
p.s. A couple other things my commit addresses:

- When no help for an identifier is found, we go to search. That search URL was being done in the back end. But now, with my commit, it consistently uses the same `raco doc` method and the `racket-documentation-search` customization variable, in the front end.

- When I added `racket-xp-mode`, the plain `racket-mode` major mode for edit buffers lost its help command. In this commit, I added it back, using the helper, as a search-for-help. As a better-than-nothing default. If someone does enable `racket-xp-mode`, then that is still used, since it's better.

So the "refactoring" I'm talking about is also about cleaning up some other loose ends, that predate your PR.

greghendershott:
p.p.s. If you (or any user) felt that <kbd>C-u C-u C-c C-d</kbd> is too much to type, for "search docs": That's a fair point. We could still add -- or even just document how the user could add their own -- command that is very thin wrapper to supply the double prefix to the main command. And they could bind that to <kbd>C-c C-s</kbd> like you had done, or whatever other key they prefer.

scolobb:
Wow, you've written quite a lot of new code!

> Before I merge: I'm wondering about "refactoring" this, both to share more implementation code, as well as to have fewer commands overall. In other words, the "helpdesk" command is "only" a variation of the main documentation command, that simply means "always search".

Completely agreed.  While I was rewriting the helpdesk commands yesterday I was thinking that there was *a lot* in common between the two, and certainly with `racket-xp-documentation`.

> Overall I like it. I'm not sure about using a double C-u C-u prefix. I like that it doesn't use up another key binding. On the other hand, does that make it too "obscure"? I don't know.

I am not particularly worried about making this functionality too obscure, once it is documented, and it is in your code.  Neither am I particularly worried about typing `C-u C-u C-c C-d`, which I personally find less cumbersome to type than `C-c RET C-a` or `C-c C-x \`.  What worries me a little is that `racket-documentation` and its sisters become harder to understand.  How do I ask it to prompt the user to enter the symbol *and* to proceed directly to the search page afterwards?  I don't think that this is possible.  In such situations I tend to split up my functions, but I don't really know how this is usually handled in Elisp.

> A couple other things my commit addresses:
> […]

Very nice!  I read your commit and I like how it streamlines the various documentation commands.  In spite of my concerns about overloading `racket-documentation`, it does feel like a natural point for centralizing the documentation functionality.

> We could still add -- or even just document how the user could add their own -- command that is very thin wrapper to supply the double prefix to the main command.

Out of curiosity, would it be something like this?

```elisp
(defun racket-helpdesk ()
  (interactive)
  (let ((current-prefix-arg '(16)))
    (racket-documentation)))
```

Maybe adding an extra function to the vanilla Racket Mode would be an overkill.  I wonder what would be the best place to place this example code snippet; in the doc strings of `racket-documentation`, `racket-xp-documentation`, and `racket-repl-documentation`?

greghendershott:
> Wow, you've written quite a lot of new code!

I think much of it was refactoring new code that _you_ wrote. As well as following up on ideas resulting from our discussion here. Thanks again!

> What worries me a little is that racket-documentation and its sisters become harder to understand. How do I ask it to prompt the user to enter the symbol and to proceed directly to the search page afterwards? 

That is the double <kbd>C-u</kbd> prefix case.

Here is the doc string for `racket-xp-documentation` (maybe easier to read here, if the git diff chopped it up):

```
"View documentation in an external web browser.

- With no prefix, uses the symbol at point.

- With a single C-u prefix, prompts you to enter a symbol.

Tries to find documentation for an identifer defined in the
current namespace.

If no such identifer exists, opens the Racket documentation
search page with the symbol. In this case, uses the variable
`racket-documentation-search' to determine whether the search is
done locally as with `raco doc`, or is done at a URL.

- With a double C-u prefix, proceeds directly to the search page.
Use this if you would like to see documentation for all
identifiers named \"define\", for example."
```

Writing doc prose is hard, so I'd definitely like to hear any suggestions you have to improve this!

---

You can just pass the prefix arg directly to the command function, I think. e.g. `(racket-xp-documentation '(4))` for the single prefix <kbd>C-u</kbd> or `(racket-xp-documentation '(16))` for the <kbd>C-u C-u</kbd> double prefix.

Maybe I could add this a tip to the "user's guide" documentation, I'm not sure.

greghendershott:
p.s. Maybe using `current-prefix-arg` is the correct thing to do, generally? I'm not sure. I'd need to read up on that aspect of Emacs. I just tried supply the `'(4)` or `'(16)` values for these own commands I've defined. It works in those cases. But if `current-prefix-arg` is better, generally, that would probably be the thing to show people in the docs.

greghendershott:
I pushed another commit: https://github.com/greghendershott/racket-mode/commit/a55b4fe5ec07cff08130e84caf365c45233f8f19

I think the `racket-documentation-search` command is the convenient target for a key binding to mean "always search". Also it makes for a good default for the <kbd>C-c C-d</kbd> binding, when `racket-xp-mode` and its `racket-xp-documentation` is not in use.

If you have any feedback, please let me know when you have a chance?

What I'd like to do: Squash all these commits from you and me, down to just a couple -- including one commit that shows your authorship of all the commits leading up to this -- and merge.

scolobb:
> What I'd like to do: Squash all these commits from you and me, down to just a couple -- including one commit that shows your authorship of all the commits leading up to this -- and merge.

I have read a55b4fe and it looks very nice.  You expanded the documentation considerably from what I initially wrote and the documentation search functions are now well streamlined.  So I completely approve the strategy you are suggesting.

> You can just pass the prefix arg directly to the command function, I think. e.g. (racket-xp-documentation '(4)) for the single prefix C-u or (racket-xp-documentation '(16)) for the C-u C-u double prefix.

Oh, OK, it's even easier than I thought!  I found the reference to `current-prefix-arg` [here](https://stackoverflow.com/questions/6156286/emacs-lisp-call-function-with-prefix-argument-programmatically), but I don't see why setting this variable would be better than explicitly calling the function with an argument.  In a functional programming mindset, supplying an argument feels like a better choice.

> Maybe I could add this a tip to the "user's guide" documentation, I'm not sure.

I think it's good to tell the users how to reduce the number of keystrokes, but the use case is probably too specific to have an explicit reference in the user's guide.  Maybe I could open and close an issue in which I ask the question and give myself an answer?  This would be still part of the documentation of the Racket Mode, but would not risk filling the user's guide with rather specific details.

greghendershott:
Thank you for that feedback!

I squashed this branch down to commit 66f892a and commit bec83a1 on a new `doc-refactor-rebase` branch. I'll merge it after I get all-green from CI....

greghendershott:
I merged those commits -- thanks!!!

scolobb:
Thank you @greghendershott for your work on this PR!  It was a real pleasure!  I'll take a tour of the open issues and see how I can help fixing them.

Issue #469, opened 2020-05-12
Auto-scaling large images


lihebi:
When working with [pict](https://docs.racket-lang.org/pict/index.html) library, I found myself often creating large images. It is slow to render in REPL. For example:

```racket
(rectangle 10000 10000)
```
takes more than 10 seconds to render. I found that the time consuming part is converting to png bytes:

```racket
(time (void (convert (rectangle 100 100) 'png-bytes)))
(time (void (convert (rectangle 1000 1000) 'png-bytes)))
(time (void (convert (rectangle 10000 10000) 'png-bytes)))
```

```
cpu time: 2 real time: 2 gc time: 0
cpu time: 100 real time: 100 gc time: 0
cpu time: 10330 real time: 10328 gc time: 522
```

Thus I wanted to **scale the image down before converting to png-bytes**. In image.rkt:

https://github.com/greghendershott/racket-mode/blob/204845665cc4ec13edc59106b102fa251458e737/racket/image.rkt#L21-L26

I would like something like:

```racket
(define (convert-image v)
  (cond
    [(pict? v) (let* ([w (pict-width v)]
                      [h (pict-height v)]
                      [factor (min (/ 640 w) (/ 480 h) 1)])
                 (save-temporary-image (convert (scale v factor) 'png-bytes)))]
    [(and (convertible? v) (convert v 'png-bytes)) => save-temporary-image]
    [else v]))
```

However, the `v` here does not seem to be `pict?`. I tried this:

```racket
(define (convert-image v) 
   (println v)
   (println (pict? v))
   (cond [(and (convertible? v) (convert v 'png-bytes)) => save-temporary-image] 
         [else v])) 
```

And it prints:

```
#<pict>
#f
```

Any ideas?

capfredf:
Can you post how `convert-image` was called?

lihebi:
This function is implemented in racket-mode. It is called just before the value of an evaluated expression is displayed:

https://github.com/greghendershott/racket-mode/blob/204845665cc4ec13edc59106b102fa251458e737/racket/print.rkt#L18-L21

For example, when you `racket-eval-last-sexp` on `(rectangle 100 100)`, that function `convert-image` will be called to generate the image, and Emacs will load the image via `insert-image`.

greghendershott:
I was just typing a similar comment. :)

[...]

PR #470, opened 2020-05-13
convert to svg-bytes instead of png-bytes; auto-scaling large images

lihebi:
Issue #469

### DIFF
--- a/racket-custom.el
+++ b/racket-custom.el
@@ -243,6 +243,26 @@ Defaults to a regexp ignoring all inputs of 0, 1, or 2 letters."
   :safe #'booleanp
   :group 'racket-repl)
 
+(defcustom racket-imagemagick-props nil
+  "ImageMagick property list given to `create-image'.
+
+For example, to scale images whose width is larger than 500
+pixels, supply (:max-width 500).
+
+This is ignored unless `racket-images-inline' is t and Emacs is
+built with with ImageMagick support."
+  :tag "ImageMagick Props"
+  :type '(plist :key-type symbol
+                :value-type (choice number string))
+  :options '((:max-width integer)
+             (:max-height integer)
+             (:background string)
+             (:width integer)
+             (:height integer)
+             (:rotation float))
+  :risky t
+  :group 'racket-repl)
+
 (defcustom racket-images-keep-last 100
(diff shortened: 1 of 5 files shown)


racket-repl.el
```diff
@@ -700,7 +700,13 @@ A value for the variable `comint-output-filter-functions'."
           (delete-region begin end)
           (goto-char begin)
           (if (and racket-images-inline (display-images-p))
-              (insert-image (create-image file) "[image]")
+              (insert-image (if (> racket-images-max-width 0)
```
greghendershott, line 703: Should we also check `(image-type-available-p 'imagemagick)` here?

racket-custom.el
```diff
@@ -243,6 +243,13 @@ Defaults to a regexp ignoring all inputs of 0, 1, or 2 letters."
   :safe #'booleanp
   :group 'racket-repl)
 
+(defcustom racket-images-max-width -1
```
greghendershott, line 246: Maybe instead default to `0`?

racket-custom.el
```diff
@@ -243,6 +243,13 @@ Defaults to a regexp ignoring all inputs of 0, 1, or 2 letters."
...
+  "The max width of inlined images. Negative values disable auto-scaling."
```
greghendershott, line 247: Maybe add a second line, something like "Note: This option only works when Emacs is built with ImageMagick (libMagickWand) support." I'm just using the phrase from the Emacs Lisp Info manual.

greghendershott:
Thanks! This looks great. I had just a few small questions, inline comments above.

greghendershott:
Thanks for making those changes!

I pushed a few more commits.

What do you think of those?

Do they work for you -- if you set the customization now named `racket-imagemagick-props` to the value `'(:max-width 1000)`)?

lihebi:
Thanks. Just tested, that works great for me.

greghendershott:
Great. I squashed down to one commit. I'll merge to `master` after I get all green from Travis CI.

greghendershott:
@timjervis Oh. Sorry! You shouldn't need to change your Racket programs. I'll open an issue for this, and `@` you and @lihebi.

timjervis:
Thanks!

PR #479, opened 2020-06-12
Fix a missing closing parenthesis in the doc

rcherrueau:
This commit adds a missing parenthesis in the documentation of `racket-xp-mode`.

Prior to this commit, I regenerated the `racket-mode.texi` following the recommendation in [CONTRIBUTING.md](https://github.com/greghendershott/racket-mode/blob/29afd2544a2f14bf29f3e542f6177579d0ae581f/CONTRIBUTING.md#docgenerateel). But the newly generated `racket-mode.texi` was full of modifications not related to my parenthesis insertion. So I reverted it back to its initial state and simply edited it manually to only add the missing closing parenthesis.

Is the manual edition good for you, or you prefer me to amend this commit with the generated `racket-mode.texi`?

### DIFF
--- a/doc/racket-mode.texi
+++ b/doc/racket-mode.texi
@@ -1049,7 +1049,7 @@ and/or slow, in your @code{racket-xp-mode-hook} you may disable them:
 	  (lambda ()
 	    (remove-hook 'pre-redisplay-functions
 			 #'racket-xp-pre-redisplay
-			 t))
+			 t)))
 @end lisp
 
 The remaining features discussed below will still work.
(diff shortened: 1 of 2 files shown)


greghendershott:
I'm sorry to be slow to look at this. Thank you for catching this!  I'll merge.

p.s. The extra changes probably resulted from you using a newer version of Emacs than I'm currently using to generate the `.texi`. For such a simple change, the manual edit is perfect. (Otherwise, I'd have suggested just committing the `.el` doc string source change, and I'd add a commit generating the `.texi. In fact, I'll do that anyway, to regenerate the HTML file used on https://racket-mode.com, and deploy it. But that's 100% OK.)

greghendershott:
p.p.s. Actually I didn't need to add a commit, just locally do a `make clean docs deploy` in the `doc` subdirectory.

Thanks again!

PR #486, opened 2020-08-04
Add support for `project.el' in `racket-repl-buffer-name-project'

yilinwei:
`project.el` is bundled with emacs by default; this adds support for it.

I've used an `fboundp`, in the case that someone is using an older emacs version or
has specifically unbound the symbol.


greghendershott:
Merged. Thanks!

PR #515, opened 2021-01-07
Add clause to the update section for Emacs newbies

soegaard:
For those of use, who only use Emacs for `racket-mode`: we a need a reminder on how to upgrade a package ;-)

### DIFF
--- a/doc/racket-mode.org
+++ b/doc/racket-mode.org
@@ -78,6 +78,10 @@ You should probably also exit and restart Emacs.
 
 ** Update
 
+In Emacs use `M-x list-packages` to list your installed packages.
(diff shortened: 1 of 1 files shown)


greghendershott:
Thanks for the PR! I agree this would be helpful!

I was thinking another way to address this would be commit 7628544. Do you have any opinions about that? If so let me know; would welcome your feedback. Otherwise I may close this PR and merge that commit instead.

greghendershott:
Thanks for your thumbs up on that commit. Merged that and deployed: https://www.racket-mode.com/#Update

Closing this. Thanks again!

PR #525, opened 2021-03-11
add the keywords define-generics and define/generic

capfredf:


capfredf:
@greghendershott I don't understand why the last testsuite failed.

greghendershott:
It looks like Emacs 28.0.50 snapshot is doing more "lint" kinds of warnings and I treat warnings as errors for the byte compile check.

I pushed a couple commits to fix that, now your commit is fine, and I merged it. Thanks!

PR #526, opened 2021-03-25
add: missing essential math notation from DrRacket

cybersyn-io:
there are some symbols that are essential to formal mathematics available in DrRacket but not Racket-Mode, I've added them, perhaps someone else will be looking for them in the future!


greghendershott:
Thanks!

Issue #476, opened 2020-06-03
browse-url doesn't handle URL anchors on macOS or Windows


sorawee:
### Steps to reproduce:

1. Paste `#lang racket add1`
2. Move the point to `add1`
3. Run `racket-xp-documentation`

### Expected behavior:

A browser shows up with the documentation

### Actual behavior

No browser shows up.

### Discussion

The problem is that the doc URL, such as `file:///Applications/Racket v7.7/doc/reference/generic-numbers.html#(def._((quote._~23~25kernel)._add1))`, has `file://` with an anchor (`#`). To quote https://www.emacswiki.org/emacs/BrowseUrl#toc23:

> Rather, it seems to be the case that `file:///C:/index.html#test` is a nonstandard URI; RFC 1630 and RFC 1738 don’t specify a `#target` (i.e., an anchor) for file URIs. (External URIs work!) Most browsers will happily parse such URIs anyway, but with Window’s `ShellExecute` the nonstandard URI is converted to a standard one before reaching the browser, and so the `#target` is stripped off.

The same problem happens in Mac. In particular, `browse-url` uses `open` to open files, and `open` has this very limitation.

This issue also came up at https://github.com/dash-docs-el/helm-dash/issues/36 and https://apple.stackexchange.com/questions/194501/open-a-file-url-with-anchor-from-terminal.

My workaround for now is to do the following:

```
  (defun my-browse-url (url &optional new-window)
    (interactive (browse-url-interactive-arg "URL: "))
    (start-process "chrome" nil "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" url))
  (setq browse-url-browser-function 'my-browse-url)
```

<details>
```
((alist-get 'racket-mode package-alist)
 #s(package-desc racket-mode
                 (20200526 1552)
                 "Racket editing, REPL, and more"
                 ((emacs
                   (25 1))
                  (faceup
                   (0 0 2))
                  (pos-tip
                   (20191127 1028)))
                 nil nil "/Users/sorawee/.emacs.d/elpa/26.3/develop/racket-mode-20200526.1552"
                 ((:url . "https://www.racket-mode.com/")
                  (:maintainer "Greg Hendershott")
                  (:authors
                   ("Greg Hendershott"))
                  (:commit . "29afd2544a2f14bf29f3e542f6177579d0ae581f"))
                 nil))
((emacs-version "26.3")
 (system-type darwin)
 (x-gtk-use-system-tooltips UNDEFINED)
[...]

PR #530, opened 2021-04-10
Make racket-browse-url functions compatible with eww

Fackelmann:
Emac's `eww` browser opens urls such as `/path/to/a/file.html` as `https://path/to/a/file.html`. This causes the documentation functions not to open the correct documentation page (such address doesn't exist).

Instead, the url provided to `eww` must be in the form `file:///path/to/a/file.html`. I have manually tested that this format is also compatible with `browse-url-default-macosx-browser`.

This PR modifies the argument in the `browse-url` call in `racket-browse-url-using-temporary-file` to use `url` (which already has the correct format) instead of `file`.

### DIFF
--- a/racket-browse-url.el
+++ b/racket-browse-url.el
@@ -53,7 +53,7 @@ customize."
          (file (make-temp-file "racket-browse-url-" nil ".html"))
          (html (format "<html><head><meta http-equiv=\"refresh\" content=\"0;url=%s\" /></head></html>" url)))
     (write-region html nil file nil 'no-wrote-file-message)
-    (browse-url file)))
+    (browse-url url)))
 
 (provide 'racket-browse-url)
 


greghendershott:
Thanks for the PR!

This seems good! Before merging, I just want to find time in the next day or two to reload my brain with the history of #476 (which was the motivation for adding the code modified by the PR) to make sure.

greghendershott:
Merged -- thanks!

greghendershott:
@Fackelmann I'm sorry, I messed up here. I said I needed to reload my brain --- but then didn't before merging it.

My fault, not yours.

I just un-merged this.

Why: Effectively, this changed `racket-browse-url-using-temporary-file` from doing what it says, into being just `browse-url`. This breaks the default case of using an external web browser on all operating systems.

Its doc string:

```elisp
(defun racket-browse-url-using-temporary-file (url &rest _args)
  "Browse a URL via a temporary HTML file using a meta redirect.

A suitable value for the variable `racket-browse-url-function'.

Racket documentation URLs depend on anchors -- the portion of the
URL after the # character -- to jump to a location within a page.
Unfortunately on some operating systems and/or versions of Emacs,
the default handling for browsing file URLs ignores anchors. This
function attempts to avoid the problem by using a temporary HTML
file with a meta redirect as a \"trampoline\".

Although the intent is to provide a default that \"just works\",
you do not need to use this. You can customize the variable
`racket-browse-url-function' instead to be `browse-url', or
`browse-url-browser-function' in case have have customized that,
or indeed whatever you want."
```

TL;DR: Awhile ago I added a customization variable, `racket-browse-url-function`. I think you want to change its value from `racket-browse-url-using-temporary-file` to just `browse-url`. Does that work for you, using eww?

greghendershott:
p.s. If setting `racket-browse-url-function` to `browse-url` doesn't quite work for you, then I'm happy to help you figure out what does work.

p.p.s. If not simply `browse-url`, maybe such a function (named `racket-browse-url-using-eww` ?) should even be added to Racket Mode as a convenience? Or maybe it's not a common enough use case? Not sure. Anyway, happy to discuss.

capfredf:
@greghendershott I think my PR should be able to make `eww` at least open the file

greghendershott:
@capfredf Oh great; I'll look at your PR #542.

Fackelmann:
Ah, my bad. I should have read better the docstring; I apologize for the confusion. I took a look at @capfredf PR and yes, that should fix this issue as well.

PR #533, opened 2021-05-02
racket-mode.org: use one line gap between header and body

utkarsh181:
Hi,

Please checkout the inconsistent space between 'Configure' header and rest of it's subheadings:

racket_doc

This irregularity or inconsistency is because Org's Texinfo exporter is getting
confused about amount of space needed between a header and
sub-headers.  To make life easier for our exporter we have to maintain
consistent distance between our headers (starting with *) and body in
org file which this pull request tries to do.

Btw, thank you for developing and maintaining racket major mode for Emacs.


greghendershott:
Thanks for the PR. I had rebased this and merged it as commit 4d23924 awhile ago, do manually closing the PR now.

Issue #552, opened 2021-08-10
Remote (tramp) racket session


fakedrake:
I do a lot of remote editing with tramp and I would like racket to be able to detect that and run on the remote machine. This could be as simple as replacing `make-process`  with `tramp-sh-handle-make-process` in `racket--cmd-open` and `:stderr` with something that tramp can handle but it probably isn't. The probem I had with the approach I propose is that `tramp-sh-handle-make-process` does not like to see pipe processes as `:stderr` keys but I don't know where to report that.

At any rate, if someone familiar with `racket-mode` code could take a look at this it would be great.

fakedrake:
I tried some stuff, I changed `racket--cmd-open` a bit. Now it looks like I need to change `make-comint-in-buffer` to look at the remote rather than the local host.

```

(defcustom racket--rkt-remote-source-dir
  "/tmp/racket-mode-source/"
  "Path to dir of our Racket source files in the remote host. The
actual remote host is determined at the calling site. See
`racket--rkt-get-source-dir'.")

(defun racket--rkt-get-source-dir (path)
  "Using PATH to determine the rempote host copy
`racket--rkt-source-dir' to `racket--rkt-remote-source-dir' in
the rempote host and return the latter.

If PATH is a local path then just retuirn
`racket--rkt-source-dir'. If PATH is nil we assume we are
operating locally and `racket--rkt-source-dir' is returned.

The return value of this function is to be used as an argument to
the racket repl invocation so it will have no tramp prefix. "
  (if (or (null path) (not (file-remote-p path)))
      racket--rkt-source-dir
    (let ((remote-file-struct (tramp-dissect-file-name path)))
      (setf (tramp-file-name-localname remote-file-struct)
            racket--rkt-remote-source-dir)
      (let ((remote-dir-name (tramp-make-tramp-file-name remote-file-struct)))
        (unless (tramp-handle-file-directory-p remote-dir-name)
          ;; COPY-DIRECTORY likes to create symlinks when the source
          ;; is a symling (eg straight keeps package repos in a
          ;; symlinked dir). This will never work when copying to a
          ;; remote host. Fortunately we can replace every symlink
          ;; create with a file copy.
          (flet ((make-symbolic-link (src dst x) (copy-file src dst t nil)))
           (copy-directory racket--rkt-source-dir remote-dir-name
                           nil t t))))
      racket--rkt-remote-source-dir)))

(defun racket--cmd-open ()
[...]

PR #553, opened 2021-08-10
Added support for remote racket

fakedrake:
When a racket session is created in a file visiting a remote server
via tramp the racket process is created in the remote server. If this
is the case the racket code running in the server is copied to the
remote server (if it is not there already). Said code is also modified
to accept connections from externa sources (security concern?).

This  is my solutiuon to issue #552 .

### DIFF
--- a/racket-cmd.el
+++ b/racket-cmd.el
@@ -77,39 +77,74 @@ that Racket Mode can find its own run.rkt file.")
   "A value we give the Racket back-end when we launch it and when we create REPLs.
 See issue #327.")
 
+
+(defun racket--rkt-get-source-dir (path)
+  "Using PATH to determine the rempote host copy
+`racket--rkt-source-dir' to `racket--rkt-remote-source-dir' in
+the rempote host and return the latter.
+
+If PATH is a local path then just retuirn
+`racket--rkt-source-dir'. If PATH is nil we assume we are
+operating locally and `racket--rkt-source-dir' is returned.
+
+The return value of this function is to be used as an argument to
+the racket repl invocation so it will have no tramp prefix. "
+  (if (or (null path) (not (file-remote-p path)))
+      racket--rkt-source-dir
+    (let ((remote-file-struct (tramp-dissect-file-name path)))
+      (setf (tramp-file-name-localname remote-file-struct)
+            racket--rkt-remote-source-dir)
+      (let ((remote-dir-name (tramp-make-tramp-file-name remote-file-struct)))
+        (unless (tramp-handle-file-directory-p remote-dir-name)
+          ;; COPY-DIRECTORY likes to create symlinks when the source
+          ;; is a symling (eg straight keeps package repos in a
+          ;; symlinked dir). This will never work when copying to a
+          ;; remote host. Fortunately we can replace every symlink
+          ;; create with a file copy.
+          (flet ((make-symbolic-link (src dst x) (copy-file src dst t nil)))
+           (copy-directory racket--rkt-source-dir remote-dir-name
+                           nil t t))))
+      racket--rkt-remote-source-dir)))
+
 (defun racket--cmd-open ()
   ;; Avoid multiple processes/buffers like "racket-process<1>".
   (racket--cmd-close)
   ;; Give the process buffer the current values of some vars; see
   ;; <https://github.com/purcell/envrc/issues/22>.
-  (cl-letf* (((default-value 'process-environment) process-environment)
-             ((default-value 'exec-path)           exec-path))
-    (make-process
-     :name            racket--cmd-process-name
-     :connection-type 'pipe
-     :noquery         t
-     :coding          'utf-8
-     :buffer          (get-buffer-create (concat " *" racket--cmd-process-name "*"))
-     :stderr          (make-pipe-process
-                       :name     (concat racket--cmd-process-name "-stderr")
-                       :buffer   nil
-                       :noquery  t
-                       :coding   'utf-8
-                       :filter   #'racket--cmd-process-stderr-filter
-                       :sentinel #'ignore)
-     :command         (list racket-program
-                            (funcall racket-adjust-run-rkt racket--run.rkt)
-                            "--auth"
-                            (setq racket--cmd-auth (format "token-%x" (random)))
-                            (if (and (boundp 'image-types)
-                                     (fboundp 'image-type-available-p)
-                                     (or (and (memq 'svg image-types)
-                                              (image-type-available-p 'svg))
-                                         (and (memq 'imagemagick image-types)
-                                              (image-type-available-p 'imagemagick))))
+  (let ((cmd (list racket-program
+                   (funcall racket-adjust-run-rkt
+                            (expand-file-name "main.rkt"
+                                              (racket--rkt-get-source-dir
+                                              default-directory)))
+                   "--auth"
+                   (setq racket--cmd-auth (format "token-%x" (random)))
+                   (if (and (boundp 'image-types)
+                            (fboundp 'image-type-available-p)
+                            (or (and (memq 'svg image-types)
+                                     (image-type-available-p 'svg))
+                                (and (memq 'imagemagick image-types)
+                                     (image-type-available-p 'imagemagick))))
                                 "--use-svg"
-                              "--do-not-use-svg"))
-     :filter          #'racket--cmd-process-filter)))
+                     "--do-not-use-svg"))))
+    (cl-letf* (((default-value 'process-environment) process-environment)
+               ((default-value 'exec-path)           exec-path))
+      (message "Running racket: %s" cmd)
+      (let* ((buf (get-buffer-create (concat " *" racket--cmd-process-name "*")))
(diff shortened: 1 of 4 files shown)


racket-repl.el
```diff
@@ -526,79 +527,98 @@ WHAT-TO-RUN may be nil, meaning just a `racket/base` namespace."
...
+          ;; (e.g. `comint-input-ring' would lose input history).
+          (unless (eq major-mode 'racket-repl-mode)
+            (when noninteractive
+              (princ "{racket--repl-start}: (racket-repl-mode)\n"))
+            (racket-repl-mode)))
+      (file-error
+       (let ((kill-buffer-query-functions nil)
+             (kill-buffer-hook nil))
+         (kill-buffer)) ;don't leave partially initialized REPL buffer
+       (message "Could not connect to REPL server: %s"
```
greghendershott, line 619: I think this is the format string the bytecode compiler is complaining about? https://github.com/greghendershott/racket-mode/actions/runs/1117388372

racket-cmd.el
```diff
@@ -77,39 +77,74 @@ that Racket Mode can find its own run.rkt file.")
...
-                            (funcall racket-adjust-run-rkt racket--run.rkt)
-                            "--auth"
-                            (setq racket--cmd-auth (format "token-%x" (random)))
-                            (if (and (boundp 'image-types)
-                                     (fboundp 'image-type-available-p)
-                                     (or (and (memq 'svg image-types)
-                                              (image-type-available-p 'svg))
-                                         (and (memq 'imagemagick image-types)
-                                              (image-type-available-p 'imagemagick))))
+  (let ((cmd (list racket-program
```
greghendershott, line 114: This assumes that `racket-program` will work on both local and remote -- e.g. that `racket` is good on both, as opposed to `/path/to/specific/version/of/racket` on one but not the other. That assumption will probably be fine for most people, so maybe it's fine to punt on this, as opposed to adding Yet Another Customization Variable? What do you think?

racket-custom.el
```diff
@@ -138,6 +138,15 @@ and `racket-repl-documentation' should look for the search page.
   :safe (lambda (val) (or (stringp val) (eq val 'local)))
   :group 'racket)
 
+(defcustom racket--rkt-remote-source-dir
```
greghendershott, line 141: Until I read your out-of-band comment, it wasn't clear to me (I was dense I guess) that this feature would involve copying the Racket Mode back end Racket code to the remote server. So I think I would probably rewrite the prose for this doc string, to make that clearer for end users. But that's something I'm happy to do myself in a separate commit.

racket/repl.rkt
```diff
@@ -128,8 +128,8 @@
 (define listener
   (tcp-listen 0  ;choose port dynamically, a.k.a. "ephemeral" port
               64
-              #f ;reuse? not recommended for ephemeral ports
-              "127.0.0.1"))
+              #t ;reuse? not recommended for ephemeral ports
```
greghendershott, line 131: Regarding the security concern: I think it might be fine if the remote back end is started using a newly added command-line argument such as `--listen-to` (defaulting to 127.0.0.1) and that value is used here. Unfortunately I guess one flaw to supplying the "what is my IP?" value automatically is the usual complication of NATs. So not yet sure best thing to do here.

greghendershott:
I made some inline comments. I hope they're helpful!

One more comment that I didn't know where to put inline: Probably this will only work reliably, especially when working with a mix of local and remote source files, when each source file has its own unique REPL session.

That is, when the value of `racket-repl-buffer-name-function` is [`racket-repl-buffer-name-unique`](https://www.racket-mode.com/#racket_002drepl_002dbuffer_002dname_002dunique). (Which is not the default value. Instead by default a single REPL is shared, which is the pre-multi-REPLs behavior.)

I think that's fine; we'd just need to document that value needs to be used. (I'd be happy to write that prose.)

racket-cmd.el
```diff
@@ -77,39 +77,74 @@ that Racket Mode can find its own run.rkt file.")
...
-     :filter          #'racket--cmd-process-filter)))
+                     "--do-not-use-svg"))))
+    (cl-letf* (((default-value 'process-environment) process-environment)
+               ((default-value 'exec-path)           exec-path))
+      (message "Running racket: %s" cmd)
+      (let* ((buf (get-buffer-create (concat " *" racket--cmd-process-name "*")))
+             (errbuf (get-buffer-create (concat " *" racket--cmd-process-name ":stderr*")))
+             (is-remote (and default-directory (file-remote-p default-directory)))
+             (make-process-fn
+              (if is-remote #'tramp-sh-handle-make-process #'make-process))
```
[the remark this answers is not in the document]
fakedrake: You are probably right, good suggestion. I didn't test it with older emacsen.

racket/repl.rkt
fakedrake: That would be nice, we need a way to find out the IP from elisp

racket-custom.el
fakedrake: You are right

greghendershott:
Please enjoy your vacation and ignore the rest of this comment until you're back!

---

Is your vacation over? Really? :smile:

You:

> Per-project repl would be nice.

That does already exist.

Me, earlier:

> One more comment that I didn't know where to put inline: Probably this will only work reliably, especially when working with a mix of local and remote source files, when each source file has its own unique REPL session.
>
> That is, when the value of racket-repl-buffer-name-function is racket-repl-buffer-name-unique. (Which is not the default value. Instead by default a single REPL is shared, which is the pre-multi-REPLs behavior.)
>
> I think that's fine; we'd just need to document that value needs to be used. (I'd be happy to write that prose.)

Actually it's not quite this simple.

A mix of local and remote files could still cause problems, even with a dedicated REPL for each.

Racket Mode expects there to be only a _single_ back end. That single back end supports zero or more REPLs.

The back end is used for quite a few commands that don't involve a REPL.

It's not clear to me what it would mean for Racket Mode to be able to juggle multiple back end processes, at all, much less a mix of local and remote.

Initially, that sounds like a buggy hairball. But, while you continue to enjoy your vacation, maybe I'll mull that over.

(Worst case, we could just advise people: "Do `M-x racket-stop-back-end` when switching between local and remote files". That's jankier than I usually like, but maybe it would be acceptable-ish (??) for your workflow and that of other people using this feature.)

fakedrake:
So one more problem with this PR is that it does not play well with `racket-open-require-path`.

greghendershott:
Over the past week, I've been working on a branch to support multiple back ends (any mix of local and remote). As well as incorporating the gist of your PR. I've been encountering a number of issues, and trying to tackle them one by one.

This has been somewhat slow going as I've been marinading myself in how tramp works etc.

> So one more problem with this PR is that it does not play well with `racket-open-require-path`.

- Yep. That directly invokes `racket-program` locally. I think it could be adapted instead to do that remotely.

- A similar issue is some of the documentation commands do a `raco doc` locally. Invoking that remotely isn't exactly the answer as that will try to open a browser remotely.

- Also there are a couple things like `racket-xp-mode` where `check-syntax` returns pathnames -- for documentation or jump-to-definition. When run remotely, those will be paths on the remote. The commands would need to change those into tramp filenames usable locally.

greghendershott:
I've continued to work on this in the [`multi-back-end`](https://github.com/greghendershott/racket-mode/tree/multi-back-end). Although I suspect there are more problems I don't yet know about, I've addressed the ones listed above -- except for invoking `raco doc` remotely.

I don't know if you're willing or able to try that branch, but if so, and if you have any comments, please let me know.

greghendershott:
Note: Working on this over several weeks, I accumulated some changes for multi/remote back ends, as well as to enhance `racket-describe`.

Why the mix: As I was dog-fooding this, I realized that `racket-describe` was important for using remote back ends, and, leaning on that more meant it should be improved. It was easiest to make the improvements during the experience of working with a remote back end, i.e. while dog-fooding.

However the `multi-back-ends` branch got into a state where it had dozens of commits, mixing both themes. This was messy. Also I wanted to the option to merge the `racket-describe` changes, sooner.

So recently I split this out:

https://github.com/greghendershott/racket-mode/tree/better-describe is squashed commits only about improving `racket-describe`. I expect to merge this first.

https://github.com/greghendershott/racket-mode/tree/rebased-multi-back-end is squashed and has only the changes related to multi back ends. I expect to merge this afterwards; periodically it needs to rebased onto `better-describe` (with the idea that `better-describe` will be merged to `master` first).

greghendershott:
To update this for anyone else following along:

- The `better-describe` branch was merged some months ago.
- Although I had set it aside for some weeks (mainly to work on the `hash-lang` branch to keep up with some developments from Racket core team in that area) I've continued to work on the `rebased-multi-back-end` branch, and @fakedrake has continued to use it and offer feedback. I think it is "feature complete".
- One important point (to me) is that if someone isn't using multiple back ends, they don't need to do any configuration. That simple scenario is supposed to "just work" like before; i.e. "do no harm" for those who don't need or care about this.
- I think `rebased-multi-back-end` might be ready to merge, depending on any more feedback.

PR #566, opened 2021-10-17
Highlight some self-eval constants and numeric literals.

NoahStoryM:
Highlight for these cases:
  1. `#T` and `#F`
  2. single-precision number and extflonum (including inf and nan)
  3. complex for `#x`, `#b`, and `#o`
  4. case insensitive

### DIFF
--- a/racket-font-lock.el
+++ b/racket-font-lock.el
@@ -88,8 +88,8 @@
          (seq symbol-start
               (or "#t" "#T" "#true" "#f" "#F" "#false"
                   (seq (any "-+")
-                       (or "inf" "inF" "iNf" "iNF" "Inf" "InF" "INf" "INF"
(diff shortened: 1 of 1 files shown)


greghendershott:
Thanks for updating the test examples and for simplifying that regular expression!

What next? Do you want me to wait while you review it or "dog-food" more yourself? Or do you think it's ready to merge?

NoahStoryM:
> What do you think?

I totally agree, it sounds really good to me.

> One wrinkle: What is '#t a.k.a. (quote #t)? That evaluates to #t, not to '#t. In other words it is not a quoted symbol. Maybe it would suffice for the regexp should exclude things starting with # as not being "quote symbols".

I have a dump way: move the code used to highlight Symbol to the end of `racket-font-lock-keywords-1`, so that other constant situations will be highlighted first.

NoahStoryM:
I fully understand how annoying it is to trade off under edge cases. Thanks for your contribution!

PR #576, opened 2021-11-06
[WIP]: Correct highlighting `define-syntaxes` and `define-values-for-syntax`.

NoahStoryM:


NoahStoryM:
The current highlighting method seems to have some problem:

```
(define-values (1st-var 2nd-var)      (values 1 2))
(define-values (1st-var second-var)   (values 1 2))
(define-values (first-var 2nd-var)    (values 1 2))
(define-values (first-var second-var) (values 1 2))

```

NoahStoryM:
I think maybe we can fix the highlighting problem of `define-syntaxes` and `define-values-for-syntax` first, and then look at how to fix the highlighting in these special cases. What do you think?

greghendershott:
I'm not sure how I overlooked this. Sorry!! Thank you for submitting it.

I'll merge this to add `define-syntaxes` and `define-values-for-syntax`.

As for handling the "unnecessary" spaces inside parens:

- On the one hand, I can't argue with the logic, "if it's legal, it should highlight".

- On the other hand, I feel like it's bad style, so I'm not super motivated to make changes to accommodate it (especially if the attempt broke something else).

- On the third hand, probably part of how/why I overlooked this, is from spending so much time recently working on the `hash-lang` branch. That takes a completely different approach to highlighting and indent, giving the `#lang` lexer full control over this. Absent that there will always be bugs like this. Someday if I merge that, an optional `racket-hash-lang-mode` minor-mode would be something to enable, which would be perfect (or at least not Racket Mode's problem to make perfect :smile:).

PR #586, opened 2022-01-21
racket-util.el: racket--rkt-source-dir - const -> var

xgqt:
in case a user wants to put racket-mode's racket (source & bytecode)
files into a different directory this will provide a easier
customization path

Signed-off-by: Maciej Barć <xgqt@gentoo.org>

### DIFF
--- a/racket-util.el
+++ b/racket-util.el
@@ -93,7 +93,7 @@ When installed as a package, this can be found from the variable
 `load-file-name'. When developing interactively, get it from the
 .el buffer file name.")
 
-(defconst racket--rkt-source-dir
+(defvar racket--rkt-source-dir
   (expand-file-name "./racket/" racket--el-source-dir)
   "Path to dir of our Racket source files. ")
 


xgqt:
> Can you say a bit about the motivation -- your specific situation where you would want these stored separately?

In Gentoo the GNU Emacs's `site-lisp` and all support scripts/executables are installed into separate directories,
respectively `/usr/share/emacs/site-lisp` and `/usr/share/emacs/etc`.

xgqt:
fixed commit message

greghendershott:
Thank you for explaining.

> In Gentoo (while a pkg is installed globally via the system pkg manager) the GNU Emacs's site-lisp and all support scripts/executables are installed into separate directories, respectively `/usr/share/emacs/site-lisp` and `/usr/share/emacs/etc`.

I don't really understand the benefit of that? But it seems harmless. Assuming people update all of `/usr/share/emacs/*/racket-mode/*` together, in sync, it shouldn't cause headaches for them or for me. I'll merge.

xgqt:
> I don't really understand the benefit of that? 

Better separation and sorting I guess.

> Assuming people update all of /usr/share/emacs/*/racket-mode/* together, in sync

`racket-mode` in Gentoo is shipped in 1 pkg, so there shouldn't be any problems

Issue #393, opened 2019-07-17
Add unicode input method for turnstile characters?


emacsomancer:
The [turnstile language](https://docs.racket-lang.org/turnstile/The_Turnstile_Reference.html#%28part._.Typing_.Unicode_.Characters%29) uses six additional special unicode characters:

- \vdash → ⊢
- \gg → ≫
- \rightarrow → →
- \Rightarrow → ⇒
- \Leftarrow → ⇐
- \succ → ≻

`racket-unicode-input-method.el` looks easy enough to extend for these, but would a pull request be worthwhile in this case? Or should  `racket-mode` be kept restricted to the current set of special characters?

stchang:
FWIW, I just manually modify `racket-unicode-input-method.el`, as you mentioned. Though I agree a better solution would be nice.

@greghendershoot, do you know if this situation is a use case for a minor mode? (I know nothing about emacs modes)

capfredf:
here is how I add one turnstile characters to racket-unicode-input-method:

```elisp
;; (require 'racket-mode) or put the code below in the `:config` part if you use `use-package`
(with-temp-buffer
    (set-input-method "racket-unicode")
    (let ((quail-current-package (assoc "racket-unicode" quail-package-alist)))
      (quail-define-rules ((append . t))
                          ("vdash" ["⊢"])
                          ("^o" ["ᵒ"]))))
```

emacsomancer:
@capfredf That also suggests a possible method of extending base `racket-mode` to allow these to be part of the `racket-mode` package itself, but leave them off by default (i.e. have a package variable like `use-turnstile-chars` which defaults to `'nil`, but when non-nil appends the turnstile special chars to the base `racket-unicode` set).

greghendershott:
- Back when I cargo-culted this from Haskell Mode, either Turnstile didn't yet exist or I wasn't aware of it. Otherwise I think I would have supplied these, too. Thanks for suggesting this!

- I'd be happy to add them, now. (Or soon-ish, after I finish tweaking things for #392 which is turning out to be a "pull the thread and the sweater unravels" thing.)

- Being Emacs, there are probably 42 ways to accomplish this. Newer Emacs have a prettify-symbols feature, for instance. I'm not immediately sure all the pros and cons, and whether Racket Mode even "needs" to supply this, anymore. (In any case I'd keep supporting it. Just not sure whether to recommend vs. deprecate it. I don't know yet.) 

[...]

PR #587, opened 2022-01-22
Use "setin " instead of "in " to produce ∈.

soegaard:
Since "in" is both a very commonly used word and a common suffix of english words, 
the replacement "in " to "∈" quickly becomes annoying when writing normal text when 
when the `racket-unicode-input-method` is enabled.

I propose to replace "in " with "setin " that way the  it is not neccessary to
keep toggling `racket-unicode-input-method` on-and-off again.

The downside is that most users will expect "in " since "\in" is used in TeX.

### DIFF
--- a/racket-unicode-input-method.el
+++ b/racket-unicode-input-method.el
@@ -142,7 +142,7 @@ can turn it off by setting `input-method-highlight-flag' to nil."
  ("vee "                ["∨"])
  ("wedge "              ["∧"])
  ("follows "            ["∘"])
- ("in "                 ["∈"])
+ ("setin "              ["∈"])
  ("ni "                 ["∋"])
  ("notin "              ["∉"])
  ("sqsubset "           ["⊏"])


greghendershott:
Thank you for the PR.

This is one of the few parts of Racket Mode I don't normally use/"dog-food" myself (or even input methods in Emacs, generally). 

So although I trust your judgment, I can't offer any good second opinion.

From issue #393 I see a few people who might care so I'll `@` them: @emacsomancer, @stchang, @capfredf 

Assuming they're on board, or don't chime in within a day or two, I'll go ahead and merge.

Thanks again!

capfredf:
+1 for this change. FWIW, I turn off the unicode input by default.  Nowadays, it is more likely I turn it on when I am writing a org document and those symbols are needed. :stuck_out_tongue_winking_eye:

sorawee:
I've never used `racket-unicode-input-method`, but I used something similar for Lean-mode. In Lean-mode however, we need to explicitly start the unicode character with `\`, similar to how LaTeX works. Couldn't the same be done here? It would also allow us to write `\in`.

bestlem:
@sorawee If you use company then the backend company-math-symbols-unicode workls like you suggest - however I don't know if it does all the things the racket addition does, it at least does the greek alphabet.

capfredf:
agda-mode also uses the same approach as lean-mode does.

greghendershott:
There are probably as many strategies and preferences for this kind of thing as there are Emacs users. :)

The `quail` / input-method approach is (IIUC):

- You type the characters. They are shown underlined while they match a substitution rule.
- Keep typing to accept.
- Or, type <kbd>C-\\</kbd> twice to toggle the input-method off and on, which cancels matching.

So you _could_ type <kbd>in</kbd> followed by <kbd>C-\ C-\\</kbd> to get plain "in" instead of "∈".  (That's probably even fewer keystrokes than typing <kbd>setin</kbd> as in this PR, depending on how you count keystrokes?)

Whether this `quail` / input-method style is more or less convenient than always needing to type a prefix like `\` ... I suppose depends on the person, and even on the particular thing they're writing from time to time?  I don't have any strong opinions anyway because I don't usually need this.

---

I suppose one thing we could do is define some var with the things most relevant to Racket -- such as the Turnstile symbols, but maybe not things like the greek letters or double-struck letters (𝔸).

And then that var might be handy for people who want to configure some other approach (like using a `\` prefix with an auto-completion framework, or whatever).

greghendershott:
@soegaard If you don't want to use "the `C-\` trick" and still want to change "in" to "setin", I'm still happy to merge this PR.

greghendershott:
p.s. I peeked at `agda-mode` and it looks like they also use `quail` / input-method. The thing is, [they have a `defcustom` which lets you tweak all of the "labels"; the default prepends `\`](https://github.com/jwiegley/agda-mode/blob/master/agda-input.el#L142).

So -- apart from probably just merging this PR as-is, i.e. fresh issue -- I could also look at adding such an option.

soegaard:
Den man. 24. jan. 2022 kl. 16.59 skrev Greg Hendershott <
***@***.***>:

> @soegaard <https://github.com/soegaard> If you don't want to use "the C-\
> trick" and still want to change "in" to "setin", I'm still happy to merge
> this PR.
>
I'd like that.

Overall I am quite happy with how  racket-unicode-input-method  works.
(Except for the "in" case).

The quail / input-method approach is (IIUC):
>
>    - You type the characters. They are shown underlined while they match
>    a substitution rule.
>
>
>    - Keep typing to accept.
>
>
>    - Or, type C-\ twice to toggle the input-method off and on, which
>    cancels matching.
>
> That's very precise.

So you *could* type in followed by C-\ C-\ to get plain "in" instead of "∈".
> (That's probably even fewer keystrokes than typing setin as in this PR,
> depending on how you count keystrokes?)


The problem here is that  "in" is so common both as a word and as a suffix
that it ruins the flow in my writing.
    Examples:    in, join, win, bin, sin (sinus), main, thin, spin etc.

Even in mathematical text I write the word "in" more often than I need ∈.

For the other abbreviations like  alpha, beta, etc, there are no problems -
they aren't common suffixes.
Therefore I think the current system with no prefix \ works fine.

/Jens Axel

stchang:
Sorry for not responding earlier, but I get tripped up by the issue in this pr all the time so I support the merge!

PR #588, opened 2022-01-28
Set lisp-indent-function

nnoodle:
This fixes weird indentation behavior from Lispy.

### DIFF
--- a/racket-common.el
+++ b/racket-common.el
@@ -244,6 +244,7 @@ property whose value is STRING. The close | syntax is set by
   (setq-local font-lock-comment-start-skip ";+ *")
   ;; -----------------------------------------------------------------
   ;; Indent
+  (setq-local lisp-indent-function #'racket-indent-function)
   (setq-local indent-line-function #'racket-indent-line)
   (setq-local indent-tabs-mode nil)
   ;; -----------------------------------------------------------------


nnoodle:
My bad, this doesn't fix the indentation behavior with Lispy.

greghendershott:
Thank you for submitting the idea, anyway!

Indent can be a bit of a hairball. Especially when code makes too many assumptions what other code is doing.

My hunch is things like Lispy need to use `prog-sexp-indent`, which IIUC is the "official" way to ask a mode to indent something. Ask it to indent, accept its result as-is, and move on. I _think_ that's probably the only way to handle this that will be reliable?

---

Having said all that, I understand it's called Lispy not Xy. You could argue it's OK for it to assume lisp indentation, whatever exactly that means for various "lisp"s.

Maybe there is even some solution using `lisp-indent-function`, but set to some new function, TBD, that is not exactly `racket-indent-function`.

Although I don't have time to dig into that right now, I'll try to take a look at this.

nnoodle:
> Maybe there is even some solution using `lisp-indent-function`, but set to some new function, TBD, that is not exactly `racket-indent-function`.

Ah, thank you. I was wondering why it wasn't set.

PR #596, opened 2022-02-16
add declare-function to keep compiler happy

tgbugs:
In emacs 29 the byte compiler fails if a function is used without being
defined or having a declare-function expression associated with it. This
causes make compile to fail.

### DIFF
--- a/racket-describe.el
+++ b/racket-describe.el
@@ -24,6 +24,7 @@
 (require 'racket-visit)
 (require 'racket-scribble)
 (require 'racket-browse-url)
+(require 'racket-back-end)
 ;; Don't (require 'racket-repl). Mutual dependency. Instead:
 (declare-function 'racket--repl-session-id "racket-repl")
 (autoload         'racket--repl-session-id "racket-repl")
--- a/racket-browse-url.el
+++ b/racket-browse-url.el
@@ -18,6 +18,7 @@
 
 (require 'racket-custom)
 (require 'racket-cmd)
+(require 'racket-back-end)
 
 (defun racket-browse-url (url &rest args)
   (when url
--- a/racket-company-doc.el
+++ b/racket-company-doc.el
@@ -18,6 +18,7 @@
 
 (require 'shr)
 (require 'racket-describe)
+(require 'racket-back-end)
 
 (defun racket--company-doc-buffer (how str)
   (pcase (racket--cmd/await (racket--repl-session-id)
@@ -44,6 +45,7 @@
            (mains (racket--scribble-mains-for-anchor mains anchor))
            (dom   `(div () ,@mains))
            (dom   (racket--walk-dom dom)))
+      (ignore tramp-verbose)
       (save-excursion
         (let ((shr-use-fonts nil)
               (shr-external-rendering-functions `((span . ,#'racket-render-tag-span)))
--- a/racket-doc.el
+++ b/racket-doc.el
@@ -20,6 +20,7 @@
 (require 'racket-cmd)
 (require 'racket-custom)
 (require 'racket-util)
+(require 'racket-back-end)
 (declare-function racket--repl-session-id "racket-repl.el" ())
 
 (defun racket--doc-assert-local-back-end ()
--- a/racket-eldoc.el
+++ b/racket-eldoc.el
@@ -17,6 +17,7 @@
 ;; http://www.gnu.org/licenses/ for details.
 
 (require 'racket-cmd)
+(require 'racket-back-end)
 
 (defun racket--do-eldoc (how repl-session-id)
   (and (racket--cmd-open-p)
--- a/racket-logger.el
+++ b/racket-logger.el
@@ -20,6 +20,7 @@
 (require 'rx)
 (require 'racket-custom)
 (require 'racket-repl)
+(require 'racket-back-end)
 
 ;; Need to define this before racket-logger-mode
 (defvar racket-logger-mode-map
--- a/racket-mode.el
+++ b/racket-mode.el
@@ -128,6 +128,9 @@
     ["Start Faster" racket-mode-start-faster]
     ["Customize..." customize-mode]))
 
+(declare-function racket-call-racket-repl-buffer-name-function "racket-repl-buffer-name" ())
+(autoload        'racket-call-racket-repl-buffer-name-function "racket-repl-buffer-name")
+
 ;;;###autoload
 (define-derived-mode racket-mode prog-mode
   "Racket"
--- a/racket-profile.el
+++ b/racket-profile.el
@@ -18,6 +18,7 @@
 
 (require 'racket-repl)
 (require 'racket-util)
+(require 'racket-back-end)
 
 (defvar-local racket--profile-project-root nil)
 (defvar-local racket--profile-results nil)
--- a/racket-repl.el
+++ b/racket-repl.el
@@ -29,6 +29,7 @@
 (require 'racket-util)
 (require 'racket-visit)
 (require 'racket-cmd)
+(require 'racket-back-end)
 (require 'comint)
 (require 'compile)
 (require 'easymenu)
@@ -37,6 +38,9 @@
 (require 'rx)
 (require 'xref)
 (require 'semantic/symref/grep)
+
+(defvar racket--back-end-auth-token)
+(declare-function  racket--what-to-run-p "racket-common" (v))
 
 ;; Don't (require 'racket-debug). Mutual dependency. Instead:
 (declare-function  racket--debug-send-definition "racket-debug" (beg end))
--- a/racket-stepper.el
+++ b/racket-stepper.el
@@ -22,6 +22,7 @@
 (require 'racket-custom)
 (require 'racket-repl)
 (require 'racket-util)
+(require 'racket-back-end)
 
 ;; Need to define this before racket-stepper-mode
 (defvar racket-stepper-mode-map
--- a/racket-xp.el
+++ b/racket-xp.el
@@ -27,6 +27,7 @@
 (require 'racket-visit)
 (require 'racket-show)
 (require 'racket-xp-complete)
+(require 'racket-back-end)
 (require 'easymenu)
 (require 'imenu)
 (require 'rx)
@@ -34,6 +35,7 @@
 (require 'xref)
 
 (declare-function racket-complete-at-point "racket-edit.el")
+(declare-function racket-browse-file-url "racket-browse-url" (path anchor))
 
 ;; TODO: Expose as a defcustom? Or even as commands to turn on/off?
 ;; Also note there are really 3 categories here: 'local 'import


greghendershott:
Thanks for the PR!

The PR seems to have several categories of changes, which I'd like to tease apart.

1. Some things you commented-out in racket-common.el and racket-xp.el. I'd rather not include those?

2. Shortening two doc strings to avoid a warning. I actually just noticed this start to crop up in a CI build for 29.0.50 yesterday that failed. Seems new w/in the last couple weeks. I have a separate branch to fix this.

3. The topic of the PR: Adding many more `declare-function`. But, as far as I know, I already have full coverage here --- I already have `declare-function` for things not already `require`d directly or transitively. AFAIK that's been sufficient/correct. I get no byte-compile warnings when doing `make compile` using Emacs 29.0.50. So if Emacs has some new story about this, I want to understand the story, as well as set up the make/CI process to alert on mistakes here (currently it's reporting no problems, for me).

Thanks again!

greghendershott:
p.s. Above when I say "Emacs 29.0.50" I mean that I just did a git fetch and rebuilt it from source, an hour ago.

tgbugs:
Woah sorry, looks like there were some stale changes that snuck in. I will clean it up. I'll also take a look to see whether the failures are back. I was compiling on an Emacs that was built with `--native-comp` maybe that was the difference?

greghendershott:
> I was compiling on an Emacs that was built with --native-comp maybe that was the difference?

I was not, so that could definitely be a difference.

I'm initially surprised the warnings would be different.

Looking at some of them, it seems for native comp it is insufficient to `(require 'a)` if `a` transitively has `(require 'b)`? It seems like if I use something from `b`, I must also/directly `(require 'b)`?

Assuming that's what's going on here -- that native-comp wants immediate requires and won't settle for transitive ones -- then:
- This seems a little tedious, but, OK, fine. I just want it checked automatically. (see last bullet)
- The better resolution (IMHO) would be to `(require 'b)` than to `declare-function`. I think `declare-function` should be saved for when `require` is impossible (due to mutual requires).
- I really want some way to add this to CI. But initially AFAICT https://github.com/purcell/setup-emacs doesn't supply a way to get a native-comp build, and neither does https://github.com/jcs090218/setup-emacs-windows. :disappointed:

Caveat: All of the above is just my first reaction based on my initial (mis?)understanding. Looking forward to learning more!

racket-describe.el
```diff
@@ -25,6 +25,9 @@
 (require 'racket-scribble)
 (require 'racket-browse-url)
 ;; Don't (require 'racket-repl). Mutual dependency. Instead:
+(declare-function racket-file-name-back-to-front "racket-back-end" (file))
```
greghendershott, line 28: I wonder if the better solution here instead would be to add `(require 'racket-back-end)` above.

Without native-comp, the `(require 'racket-cmd)`, which in turn does `(require 'racket-back-end)`, suffices --- the transitive require means something.

Whereas with native-comp (if I'm guessing correctly?) it wants to see `'racket-back-end` required directly, here?

tgbugs: I think you are right. native-comp seems to want direct imports and will fail if the import of a form is transitive.

tgbugs:
I made the changes to add direct imports which seems to have solved the issue more nicely in nearly every case. There are a few exceptions and I'm not sure why those forms in particular cause issues, possibly because they are used in macros or something?

tgbugs:
Re: `--with-native-comp` see https://github.com/purcell/nix-emacs-ci/pull/49 and https://github.com/purcell/nix-emacs-ci/issues/19.

greghendershott:
> I made the changes to add direct imports which seems to have solved the issue more nicely in nearly every case.

Thanks; that seems much nicer!

> There are a few exceptions and I'm not sure why those forms in particular cause issues, possibly because they are used in macros or something?

That's interesting; I'm not sure why either.

I think it would probably be OK to just take those at face value for now.

---

p.s. For belt+suspenders I should probably also change `make compile` to run Emacs to do a `check-declare-directory`.

greghendershott:
@tgbugs I asked on https://emacs.stackexchange.com/questions/70580/transitive-requires-dont-work-with-native-comp.

The [comments so far](https://emacs.stackexchange.com/questions/70580/transitive-requires-dont-work-with-native-comp?noredirect=1#comment113561_70580): Ask on the mailing list, or, `M-x report-emacs-bug`.

Since you're an active user of native-comp, I wanted to see if you'd be interested in doing either of those follow-up steps? (Otherwise I can.)

tgbugs:
I can take a shot at the report.

greghendershott:
Thanks! Although I did try doing a native comp build from source, I don't seem to have the `libccjit` deps sorted out correctly on Ubuntu 18.04. I'll try again. Meanwhile since you already have a native-comp Emacs built, it would be great if you're willing to do `M-x emacs-bug-report` from that.

greghendershott:
To follow up on this: I eventually did arrange to build from source ("Emacs 29.0.50") today using `--with-native-compilation`.

Using that, I am **not** able to reproduce this "transitive `require`" issue, at least using this simplified recipe:

```sh
$ cat a.el 
;;; a.el
(defun a () nil)
(provide 'a)

$ cat b.el 
;;; b.el
(require 'a)
(defun b () nil)
(provide 'b)

$ cat use.el
;;; use.el
(require 'b)
(a) ;Do we get a warning about this?
(b)
(c) ;We should definitely get a warning about this

$ ~/emacs/src/emacs --batch -Q -L . --eval '(setq byte-compile-warnings t)' -l bytecomp -f batch-byte-compile use.el
# does NOT print any warning about `a`
In end of data:
use.el:5:2: Warning: the function ‘c’ is not known to be defined.
```

tgbugs:
I will continue to investigate on my end.

greghendershott:
> I will continue to investigate on my end.

Thank you!

---

When you have time, here's one other data point. Instead of the simple recipe, I try the actual Racket Mode files before this PR (that should exhibit the problem).

1. I checkout the commit just prior to the one merged in this PR:

```
$ git checkout 22e3197
```

2. And -- because I'm unsure when Emacs will native vs. byte comp; I need to learn more -- hack the `Makefile` like so to make it just native comp:

```
diff --git a/Makefile b/Makefile
index 666521b..39f049f 100644
--- a/Makefile
+++ b/Makefile
@@ -18,10 +18,9 @@ batch-emacs := $(EMACS) --batch -Q -L . --eval '(package-initialize)'
 
 byte-compile := \
   $(batch-emacs) \
-  -l bytecomp \
+  -l comp \
   --eval '(setq byte-compile-warnings t)' \
-  --eval '(setq byte-compile-error-on-warn t)' \
-  -f batch-byte-compile
+  -f batch-native-compile
 
 %.elc : %.el
        $(byte-compile) $<
```

3. And finally do:

```
$ EMACS=~/emacs/src/emacs make compile
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-back-end.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-browse-url.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-bug-report.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-cmd.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-collection.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-common.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-company-doc.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-complete.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-custom.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-debug.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-describe.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-doc.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-edit.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-eldoc.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-font-lock.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-imenu.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-indent.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-keywords-and-builtins.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-logger.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-mode.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-parens.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-ppss.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-profile.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-repl-buffer-name.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-repl.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-scribble.el

In end of data:
racket-scribble.el:55:6: Warning: the function ‘libxml-parse-html-region’ is not known to be defined.
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-show.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-smart-open.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-stepper.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-unicode-input-method.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-util.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-visit.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-wsl.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-xp-complete.el
/home/greg/emacs/src/emacs --batch -Q -L . --eval '(package-initialize)' -l comp --eval '(setq byte-compile-warnings t)' -f batch-native-compile racket-xp.el
```

I don't get any warnings about undeclared functions.  (The sole warning is about XML because I was a potato and didn't `./configure` that option correctly.)

In other words, it seems like maybe your build of Emacs differs from mine? Maybe there was some problem with transitive requires, that was later fixed?

tgbugs:
I'm going to do a rebuild and try again. For the record I was at https://github.com/emacs-mirror/emacs/commit/51e51ce2df46fc0c6e17a97e74b00366bb9c09d8.

tgbugs:
Appears to be gone now as of https://github.com/emacs-mirror/emacs/commit/9372e543a92306dabfb03a3c1f97ceb5cb4d8608, when testing against 22e319754dcf650e282b3ba33b9d0ee3cda81007. Perhaps ok to revert if desired?

greghendershott:
Thank you for trying that again!

So IIUC, it seems it was weird timing. You happened to build c. 10 days ago when there was some issue with transitive requires. That was fixed by the time I tried a couple days ago.

Glancing at the commits in that  time range, I have no idea which one was the fix. My W.A.G. is there were some changes around `current-load-list` and `load-history` that caused the native compiler not to know about transitively loaded functions... and then to know again...???

I guess I'll post something like that as a self-answer to https://emacs.stackexchange.com/questions/70580/transitive-requires-dont-work-with-native-comp ... but not mark the answer as accepted. Maybe someone else will come along and answer. It probably doesn't matter, but having raised the question there, I feel like I should create a chance for it have more value.

> Perhaps ok to revert if desired?

I'm OK with the "unnecessary" `require`s. I don't love the extra `defvar`; I might do a tiny commit to remove that, instead of reverting the whole thing.

Mainly I'm glad knowing that the non-`--with-native-compilation` builds used for local and/or CI testing are again sufficient, and are intended to be (there's no intentional semantic difference with native compilation).

Issue #598, opened 2022-02-18
Has ^M in the racket-steppper-mode buffer?


chansey97:
I am trying to use macro expander in racket-mode.

Everything is fine, except that the stepper buffer will display an extra `^M` 

For example,

```
#lang racket

(define-syntax-rule (baz (arg1 ...) (arg2 ...) ) 
  (list '(+ 42 arg1 arg2) ...))

(baz (1 2) (11 22))
```

Placing the mouse cursor at `(baz (1 2) (11 22))|` and `M-x racket-expand-last-sexp`, the stepper buffer will show:

![racket-mode](https://user-images.githubusercontent.com/4015056/154594004-8f494ea2-527c-41f4-9305-a99cec92a6dc.JPG)

Thank you for this great work.

My racket-mode version is racket-mode-20220216.1917 from elpa, operation system is windows 7, Racket v8.0 [bc].

racket-bug-report:

<details>
<pre>
((alist-get 'racket-mode package-alist))
((emacs-version "27.2")
 (system-type windows-nt)
 (x-gtk-use-system-tooltips UNDEFINED)
 (major-mode racket-mode)
 (racket--el-source-dir "c:/Users/Chansey/AppData/Roaming/.emacs.d/site-lisp/racket-mode-20220216.1917/")
 (racket--rkt-source-dir "c:/Users/Chansey/AppData/Roaming/.emacs.d/site-lisp/racket-mode-20220216.1917/racket/")
 (racket-program "C:\\Program Files\\Racket\\Racket.exe")
 (racket-command-timeout 10)
 (racket-path-from-emacs-to-racket-function UNDEFINED)
 (racket-path-from-racket-to-emacs-function UNDEFINED)
 (racket-browse-url-function racket-browse-url-using-temporary-file)
 (racket-documentation-search-location "https://docs.racket-lang.org/search/index.html?q=%s")
 (racket-xp-after-change-refresh-delay 1)
 (racket-xp-mode-lighter
  (:eval
   (racket--xp-mode-lighter)))
 (racket-xp-highlight-unused-regexp "^[^_]")
 (racket-repl-buffer-name-function nil)
 (racket-submodules-to-run
  ((test)
   (main)))
 (racket-memory-limit 2048)
 (racket-error-context medium)
 (racket-repl-history-directory "~/.emacs.d/racket-mode/")
 (racket-history-filter-regexp "\\`\\s *\\'")
 (racket-images-inline t)
 (racket-imagemagick-props nil)
[...]

PR #600, opened 2022-02-18
Fix macro expander has extra ^M in windows.

chansey97:
see https://github.com/greghendershott/racket-mode/issues/598

### DIFF
--- a/racket/commands/macro.rkt
+++ b/racket/commands/macro.rkt
@@ -158,7 +158,9 @@
   (define after-file  (make-temporary-file-with-text after-text))
   (define out (open-output-string))
   (begin0 (parameterize ([current-output-port out])
-            (system (format "diff -U ~a ~a ~a" -U before-file after-file))
+            (if (eq? (system-type) 'windows)
+                (system (format "diff --strip-trailing-cr -U ~a ~a ~a" -U before-file after-file))
+                (system (format "diff -U ~a ~a ~a" -U before-file after-file)))
             (match (get-output-string out)
               ["" " <empty diff>\n"]
               [(pregexp "\n(@@.+@@\n.+)$" (list _ v)) v]))


racket/commands/macro.rkt
```diff
@@ -158,7 +158,9 @@
   (define after-file  (make-temporary-file-with-text after-text))
   (define out (open-output-string))
   (begin0 (parameterize ([current-output-port out])
-            (system (format "diff -U ~a ~a ~a" -U before-file after-file))
+            (if (eq? (system-type) 'windows)
+                (system (format "diff --strip-trailing-cr -U ~a ~a ~a" -U before-file after-file))
```
greghendershott, line 162: This assumes that all `diff` utilities someone might use on Windows, have such a `--strip-trailing-cr` command-line flag. Is that a safe assumption?

racket/commands/macro.rkt
```diff
@@ -158,7 +158,9 @@
...
+                (system (format "diff -U ~a ~a ~a" -U before-file after-file)))
             (match (get-output-string out)
```
greghendershott, line 164: Another idea would be to change the result string ourselves. For example here do something like `(match (regexp-replace* #rx"\r\n" (get-output-string out) "\n")`.

I think it's probably fine to do this unconditionally (it's harmless when not on Windows)?

greghendershott:
Thank you very much for the pull request!

racket/commands/macro.rkt
chansey97: Windows 7 has no `diff` at all. I am using cygwin, so lucky to have it. MinGW also has this `diff` with `--strip-trailing-cr`. So I think it is safe.

PS. I don't know windows 10 or windows 11. I think they have not `diff` as well, because windows has other tool, called `fc` or `windiff`.

racket/commands/macro.rkt
chansey97: I have tried your approach, it is OK. I also think it should be harmless.

PR #643, opened 2022-11-03
The tag of here string can contain spaces.

NoahStoryM:


racket-common.el
```diff
@@ -87,7 +87,7 @@
     ;; char syntax around the "body" so it's treated as a string for
     ;; indent, nav, font-lock. Think of the \n in #<<ID\n as the open
     ;; | quote and the \n in ^ID\n as the close | quote.
-    ((rx "#<<" (group (+? (not (any blank ?\n)))) (group ?\n))
+    ((rx "#<<" (group (+? (not ?\n))) (group ?\n))
```
greghendershott, line 90: This `not` syntax doesn't work on some older versions of Emacs that Racket Mode still supports, which is why the automated tests are failing for those.

Instead of `(not ?\n)` I think you need `(not (any ?\n))`.

racket-font-lock.el
```diff
@@ -43,7 +43,7 @@
       ;;
       ;; We only handle the opening #<<ID here. The remainder is
       ;; handled in `racket-font-lock-syntactic-face-function'.
-      (,(rx (group "#<<" (+? (not (any blank ?\n)))) ?\n)
+      (,(rx (group "#<<" (+? (not ?\n))) (group ?\n))
```
greghendershott, line 46: Here also, instead of `(not ?\n)` I think you need `(not (any ?\n))`.

PR #653, opened 2023-02-08
builtins: update keyword and builtin lists for Racket 8.8.0.6

Bogdanp:
I noticed the `for/list/concurrent` forms weren't being highlighted so I went ahead and updated the lists. The last update was circa 8.5.


greghendershott:
Thanks! I'll merge.

p.s. As for indent: AFAICT that already works fine for the for new `for/lists/concurrent` forms. Because we have a regexp check for `for/` or `for*/` forms. That indent handler accepts optional keywords like `#:length` or `#:group` (as well as Typed Racket annotations if any).

PR #670, opened 2023-10-07
Explicitly font-lock-fontify-buffer after re-enabling font-lock-mode

### LINKED ISSUE #642: racket-hash-lang-mode not highlighting buffer with font-lock
I’m trying out the head of the `hash-lang` branch on Emacs 28.1 with Racket 8.6. (I know it’s pre-release, so you might already know about this problem.) It works and the buffers are definitely talking to the backend, which is great! (Among other things, Scribble indentation works correctly like in DrRacket.) But neither font-lock nor recognizing when the `#lang` has changed are working. Specifically, I can colour the whole buffer by `M-:` evaluating `(racket--hash-lang-font-lock-fontify-region (point-min) (point-max))`, which works fine, but it doesn’t automatically font-lock the buffer as I type, and if I start an empty buffer and add any `#lang` line, it doesn’t recognize the new language I set.

`after-change-functions` in a `racket-hash-lang-mode` buffer are `(t racket--hash-lang-after-change-hook)`, and `font-lock-fontify-region-function` is indeed `racket--hash-lang-font-lock-fontify-region`, so I don’t understand what’s up here.

dpk:
Fixes #642 

My guess is that `font-lock-set-defaults` wasn’t being called.

### DIFF
--- a/racket-hash-lang.el
+++ b/racket-hash-lang.el
@@ -166,6 +166,7 @@ A discussion of the information provided by a Racket language:
          (font-lock-mode 1)
          (read-only-mode -1)
          (cancel-timer timer)
+         (font-lock-fontify-buffer)
          (setq-local header-line-format nil))))))
 
 (defun racket--hash-lang-delete ()


greghendershott:
It looks like the Emacs byte compiler is complaining that `font-lock-fontify-buffer` is intended only as an end user interactive command; wants us to use `font-lock-flush` or `font-lock-ensure` instead. And actually in my testing now, just `font-lock-set-defaults` suffices.

Do you want to revise the PR for me to merge?

Or, do you want me to close the PR and take it from here (giving you credit in my own commit message!!)?

I suggest the latter because I've already taken so much of your time, and I can move forward. But I'm fine with either. Thanks again!

PR #704, opened 2024-03-16
racket-doc: Dont try to expand racket-program as directory

nverno:
Expanding `racket-program` expands the path in whatever the current directory is unless `racket-program` already happens to be an absolute path.

### DIFF
--- a/racket-doc.el
+++ b/racket-doc.el
@@ -58,7 +58,7 @@ Centralizes how to issue doc command and handle response correctly."
 
 (defun racket--search-doc-locally (str)
   (racket--doc-assert-local-back-end)
-  (call-process (expand-file-name racket-program)
+  (call-process racket-program
                 nil ;INFILE: none
                 0   ;DESTINATION: discard/don't wait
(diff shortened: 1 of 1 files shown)


greghendershott:
Huh, I wonder what I was thinking there.

Maybe it's copypasta from `racket-shell.el`:

```elisp
(defun racket--shell-or-terminal (args)
  (racket--save-if-changed)
  (let* ((exe (shell-quote-argument
               (if (file-name-absolute-p racket-program)
                   (expand-file-name racket-program) ;handle e.g. ~/
                 racket-program)))
         (cmd (concat exe " " args))
         (win (selected-window)))
    (funcall racket-shell-or-terminal-function cmd)
    (select-window win)))
```

That has a comment about handling things like `~`. OK. However (1) there it's nested in a `file-name-absolute-p` test, avoiding the gotcha you mention. And (2) there it's solving a problem (maybe?) that we don't have when using `call-process`, which is OK with paths like `~/racket/bin/racket`.

So I'll merge this -- thanks!

PR #717, opened 2024-09-10
Remove "oint" keybinding in racket-unicode input method

bkc39:
The `oint` quail prefix activates when writing common identifiers such as `point` and `pointer`.  This removes the binding from the input method out of convenience.

### DIFF
--- a/racket-unicode-input-method.el
+++ b/racket-unicode-input-method.el
@@ -130,7 +130,7 @@ can turn it off by setting `input-method-highlight-flag' to nil."
  ("prod "               ["∏"])
  ("coprod"              ["∐"])
  ("integrate "          ["∫"])
- ("Oint"                ["∮"])
+ ("Oint "               ["∮"])
  ("vee "                ["∨"])
  ("wedge "              ["∧"])
  ("follows "            ["∘"])
--- a/racket-unicode-input-method.el
+++ b/racket-unicode-input-method.el
@@ -130,6 +130,7 @@ can turn it off by setting `input-method-highlight-flag' to nil."
  ("prod "               ["∏"])
  ("coprod"              ["∐"])
  ("integrate "          ["∫"])
+ ("Oint"                ["∮"])
  ("vee "                ["∨"])
  ("wedge "              ["∧"])
  ("follows "            ["∘"])
--- a/racket-unicode-input-method.el
+++ b/racket-unicode-input-method.el
@@ -130,7 +130,6 @@ can turn it off by setting `input-method-highlight-flag' to nil."
  ("prod "               ["∏"])
  ("coprod"              ["∐"])
  ("integrate "          ["∫"])
- ("oint "               ["∮"])
  ("vee "                ["∨"])
  ("wedge "              ["∧"])
  ("follows "            ["∘"])


greghendershott:
Thanks for the PR!

This is one feature I don't "dog food" myself. So I don't have any useful opinion. I'm happy to defer to you, and merge this.

Comment: I didn't realize these trigger in the middle of words. That seems... tricky.

Question: Just out of curiosity, are these case sensitive? If so, would it make any sense maybe to change it from "oint" to "Oint"?

(Not seeking a long discussion. Just wanted to ask before merging.)

capfredf:
I feel like we should use a leader/prefix key to turn on the unicode input method

bkc39:
> This is one feature I don't "dog food" myself. So I don't have any useful opinion. I'm happy to defer to you, and merge this.

Yea, this came up on the Racket Discord channel and some others seem to have experienced the same issue.
> 
> Comment: I didn't realize these trigger in the middle of words. That seems... tricky.

Yea what i generally do is `C-g` right before the last character to exit the quail completion.

> Question: Just out of curiosity, are these case sensitive? If so, would it make any sense maybe to change it from "oint" to "Oint"?

The bindings are case sensitive. I would personally be fine with "Oint" as the mapping and agree that is better than removing the binding completely

bkc39:
> I feel like we should use a leader/prefix key to turn on the unicode input method

`C-\` toggles the input method on and off. The issue with `oint` is that because `oint` is a common substring of Racket identifiers you end up having to toggle a lot.

capfredf:
I was not clear. In the agda2 mode, the Agda unicode input method is turned on by default, but one needs to press \ to input the unicode symbols.

bkc39:
I would be fine with something similar to Agda mode. I also use the `TeX` input method which has the same convention as Agda, but I think that would be different enough behavior that we may want to consider making a different `quail` package altogether. `racket-escaped-unicode-input-method`?

greghendershott:
It looks like `agda2-mode` does this, not in the translation map itself, but via a user customization [`agda-input-tweak-all`](https://github.com/emacsmirror/agda2-mode/blob/master/agda-input.el#L131-L148) to tweak that map in various ways.

The customization is extremely generalized; including [predefined functions](https://github.com/emacsmirror/agda2-mode/blob/master/agda-input.el#L29-L112). (First glance, I wonder is that excessive? But subsequent glances, maybe I'd say sure why not.)

greghendershott:
In any case, _some_ customizable prefix (and/or suffix?) seems the way to go.

For example, if I were using this mode, a <kbd>\\</kbd> prefix would drive me nuts given how often I need that for escapes in Racket strings and regexps. In fact that's bound to `paredit-backslash` for me. Also, I had to type `\\` twice to escape itself here in this markdown comment. :)

greghendershott:
> It looks like `agda2-mode` does this, not in the translation map itself, but via a user customization [`agda-input-tweak-all`](https://github.com/emacsmirror/agda2-mode/blob/master/agda-input.el#L131-L148) to tweak that map in various ways.
> 
> The customization is extremely generalized; including [predefined functions](https://github.com/emacsmirror/agda2-mode/blob/master/agda-input.el#L29-L112). (First glance, I wonder is that excessive? But subsequent glances, maybe I'd say sure why not.)

This seems like the right direction to proceed.

I'd be fine cargo-cult-ing all of the agda-mode code (because, hell, originally this was mostly cargo-cult-ed from haskell-mode). Of course, just using the rackety translations, instead.

I'd also be fine limiting the predefined utility functions, for users when customizing, to just the stuff pertaining to changing the prefix. No strong opinion here either way.

@bkc39 I don't know if this is something you would still like to do as a PR?  If that feels like too much: I could take a crack at this, probably within a week after getting a finish on another issue I'm currently working on.

bkc39:
@greghendershott I could probably take a swing at this this weekend. Do you think this should be included in this PR or submitted separately? Do you think we should update `racket-unicode` input method or make another that "inherits" from `racket-unicode` like they "inherit" from `TeX` in `agda2-mode`?

I actually have the beginnings of this in my personal config but I am not sure the best way to port it into the `racket-mode` package.

greghendershott:
It seems that the existing translation table could be fine, provided there's a way users can customize a prefix or suffix?

If you do agree, then I'd say the PR could morph into or be replaced by the new "cargo cult agda2-input" plan.

But maybe you disagree, because you'd rather not type any prefix at all; you just want to change "oint" to "Oint" and you'd be happy? So for you the esting PR would suffice?

If so, that's OK, too. Just let me know. I could merge your simpler change, and then you or I could tackle the bigger change, later.

Or,

bkc39:
> It seems that the existing translation table could be fine, provided there's a way users can customize a prefix or suffix?

I agree. 

> But maybe you disagree, because you'd rather not type any prefix at all; you just want to change "oint" to "Oint" and you'd be happy? So for you the esting PR would suffice?

I think that in the immediate term the "oint" binding specifically is what I and others would like change and in that sense the existing PR would suffice to remove the pain point.

> If so, that's OK, too. Just let me know. I could merge your simpler change, and then you or I could tackle the bigger change, later.

My preference would be to merge this change and then have a separate PR for the bigger change as that is likely to warrant more scrutiny.

greghendershott:
Merged; thanks!

Issue #718, opened 2024-09-14
Use customization ideas from agda2-mode's input method


greghendershott:
As discussed in #717, it could be useful to borrow some/all of the customization "infrastructure" ideas from agda2-mode: https://github.com/emacsmirror/agda2-mode/blob/master/agda-input.el#L131-L148. This would allow users to configure a prefix, suffix, inherit from other input methods, etc.

greghendershott:
Commit 61d60c6 is a "first draft" example. Welcome anyone's feedback!

```
;; PROVENANCE: This is a copy and modification of the MIT licensed
;; <https://github.com/emacsmirror/agda2-mode/agda-input.el>. Most of
;; the modification consists of just changing "agda-" name prefixes to
;; "racket-".
;;
;; On the one hand, most of this code is applicable to defining and
;; customizing any input method. Ideally, we'd like it to exist as a
;; distinct, "generic" package -- named e.g. "quail-customize"? -- for
;; use by any input method, so we could use it that way.
;;
;; On the other hand, it doesn't exist in that form. Even if we wanted
;; to advise users to install the speific agda2-mode package (i.e. use
;; it as a "library"), they ship it via via Stack/Cabal -- not via any
;; Emacs Lisp package repo. We can't ask a typical user install it
;; that way.
;;
;; Finally, I would be happy to copy-paste this exactly, using the
;; "agda-" prefixed names. But if an Agda user happens to have it
;; installed the official way, that would conflict.
;;
;; TL;DR: Although doing a copy and rename is a bad way to reuse this
;; code, it seems the least worst way. :(
```

Potentially an even better idea would be to do this in the form of a new "quail-customize" package to be available on MELPA and/or NonGNUELPA. Although I'm not sure I have the bandwidth to do this myself, maybe someone else would??

PR #719, opened 2024-09-14
Customization of unicode input method

bkc39:
Make the keybindings customizable via user-defined functions. Bindings inherit from `racket-unicode`

this should be tested as some of the code and docs were written by LLM, but it seems to work well enough for me

```elisp
(racket-custom-unicode-add-tweaks!
 (remove-suffix " ")
 (prefix-with "\\"))
(racket-custom-unicode-input-method-enable) ;; now \alpha maps to unicode alpha
```

Partially addresses #718


greghendershott:
Unfortunately for license provenance reasons I'm unable to merge code sourced by an LLM.

I understand some (maybe even most?) you wrote by hand. That's why I wanted to mention this as soon as I noticed -- before you put much more work into it.  :disappointed: 

I really appreciate you taking the time to work on this. I feel badly. I'm definitely not suggesting you did anything wrong; just that it wouldn't work for this project.

---

In any case, I really did have in mind what I'd mentioned: Borrowing the agda2-input code almost verbatim. Because I've grown to appreciate its approach. Also the provenance would be clear.

[In fact: I wonder if we could recommend people install the agda package; it doesn't seem to depend on other packages, so it's pretty light. And we'd literally use its code, that way. And we could supply an initial configuration of it, equivalent to our status quo. Something like that.]

greghendershott:
p.s. Since I seem to have particular ideas about this, maybe it would be better if I worked on this, at least to do an overall "first draft" commit? And then I'd welcome your review/contributions against that baseline, if you wanted?

It doesn't seem fair to have you try to chase down some idea if I'm not being clear about it. And if you want to contribute using some approaches, but not anothers, of course that's valid, its up to you not me!

bkc39:
> I really appreciate you taking the time to work on this. I feel badly. I'm definitely not suggesting you did anything wrong; just that it wouldn't work for this project.

No worries. I would say maybe add this to your contributing page. I do nearly all programming with some LLM assistance these days and I'm sure I'm not the only one. If this is going to be your policy I would make that more clear to potential contributors. The thought that there might be some license issues never occurred to me.

bkc39:
> p.s. Since I seem to have particular ideas about this, maybe it would be better if I worked on this, at least to do an overall "first draft" commit? And then I'd welcome your review/contributions against that baseline, if you wanted?
> 
> It doesn't seem fair to have you try to chase down some idea if I'm not being clear about it. And if you want to contribute using some approaches, but not anothers, of course that's valid, its up to you not me!

Yea this seems good. I'll keep an eye out.

PR #737, opened 2025-01-18
Introduce a new envvar to set a longer timeout for Debian buildds

manphiz:
* Some of the Debian supported architectures are very slow and requires a longer timeout for some of the tests to finish.

### DIFF
--- a/test/racket-tests.el
+++ b/test/racket-tests.el
@@ -40,7 +40,16 @@
 (defconst ci-p (getenv "CI")
   "Is there an environment variable saying we're running on CI?")
 
-(defconst racket-tests/timeout (if ci-p 60 10))
+(defconst debian-buildd-p (getenv "DEBIAN_BUILDD")
+  "Is there an environment variable saying we're running on Debian buildd?")
+
+(defconst racket-tests/timeout
+  (cond
+   ;; Some Debian supported architectures are very slow and requires
+   ;; a longer timeout for some of the tests to finish.
+   (debian-buildd-p 900)
+   (ci-p 60)
+   (t 10)))
 
 (defun racket-tests/type (typing)
   (let ((blink-matching-paren nil)) ;suppress "Matches " messages
--- a/test/racket-tests.el
+++ b/test/racket-tests.el
@@ -269,7 +269,7 @@ c.rkt. Visit each file, racket-run, and check as expected."
 
 (ert-deftest racket-tests/profile ()
   "Exercise `racket-profile'."
-  (skip-unless (not ci-p))
+  (skip-unless (not (or ci-p debian-buildd-p)))
   (message "racket-tests/profile")
   (racket-tests/with-back-end-settings
     (let* ((path (make-temp-file "test" nil ".rkt"))
@@ -726,11 +726,11 @@ want to use the value of `racket-program' at run time."
       dur)))
 
 (ert-deftest racket-tests/indent-speed-1 ()
-  (skip-unless (not ci-p))
+  (skip-unless (not (or ci-p debian-buildd-p)))
   (should (< (racket-tests/indent-time "example/class-internal.rkt") 10)))
 
 (ert-deftest racket-tests/indent-speed-2 ()
-  (skip-unless (not ci-p))
+  (skip-unless (not (or ci-p debian-buildd-p)))
   (should (< (racket-tests/indent-time "example/core.scm") 10)))
 
 ;;; Font-lock


greghendershott:
Thanks for the pull request!

At first I didn't understand this. But digging around, I'm guessing this is this for running the Emacs Lisp tests on a [`buildd`](https://www.debian.org/devel/buildd/) instance, for the Debian package called [`racket-mode`](https://ci.debian.net/packages/r/racket-mode/)?

Can you help me understand the purpose of that Debian package?

Why I ask: Racket Mode is a combination of both (a) Emacs Lisp and (b) Racket code. It is designed for both the Emacs Lisp and Racket code to be updated in sync and delivered together as an Emacs package (or people can install both directly from source, here).

It is definitely not designed/maintained to support people getting some of the code via one method, and some via another method. [In other words, I deliberately don't supply both an Emacs Lisp package and a separate Racket package, and ask users to update both. In fact, I frequently change the API between the two, in ways that don't matter when they're delivered together -- but could break badly if delivered separately and not updated perfectly in lock step.]

At a quick glance at the Debian package's test logs, I see only the Emacs tests being run (e.g. `make test-elisp`). But I don't see the Racket tests being run (`make test-racket`). The right thing to do is run both, (`make test`).

So I'm a little concerned that the Debian package might not be doing the right thing?

Can you help me understand more? Thanks!

manphiz:
Hi Greg,

Thanks for your reply, and sorry for not being very clear about the
purpose of the PR.  Please see my replies below.

Greg Hendershott ***@***.***> writes:

> Thanks for the pull request!
>
> At first I didn't understand this. But digging around, I'm guessing
> this is this for running the Emacs Lisp tests on a
> [`buildd`](https://www.debian.org/devel/buildd/) instance, for the
> Debian package called
> [`racket-mode`](https://ci.debian.net/packages/r/racket-mode/)?
>

This is correct.

> Can you help me understand the purpose of that Debian package?
>

The Debian racket-mode package[1] basically provides the exact content
of this repo and ship as a deb package.  The result is like installing
through nongnu ELPA, but is done using Debian's apt and follows Debian
conventions.

> Why I ask: Racket Mode is a combination of both (a) Emacs Lisp and (b)
> Racket code. It is designed for both the Emacs Lisp and Racket code to
> be updated in sync and delivered together as an Emacs package (or
> people can install both directly from source, here).
>
> It is definitely not designed/maintained to support people getting
> some of the code via one method, and some via another method. [In
> other words, I deliberately don't supply both an Emacs Lisp package
> and a separate Racket package, and ask users to update both. In fact,
> I frequently change the API between the two, in ways that don't matter
> when they're delivered together -- but could break badly if delivered
> separately and not updated perfectly in lock step.]
>

Ack.  IIUC racket-mode contains Elisp code and racket code to provide
editing and debugging support for Emacs, is this understanding correct?

Ideally I would expect the racket-mode package only contains ELisp code
to interact with a stable racket interface for operations, but I guess
this may be fine before the interface stabilizes.  Is this understanding
correct?

> At a quick glance at the Debian package's test logs, I see only the
> Emacs tests being run (e.g. `make test-elisp`). But I don't see the
> Racket tests being run (`make test-racket`). The right thing to do is
> run both, (`make test`).
>

Indeed.  And actually, the Emacsen team uses a tool "dh-elpa", which is
based on Debian's debhelper, to detect and run ERT and Buttercup tests
without relying on any Makefiles or other build systems, hence only
Elisp related tests are run.

> So I'm a little concerned that the Debian package might not be doing
> the right thing?
>

Or doing half of the work: yes, currently the racket test is not run at
the moment, and personally I think having this coverage is better than
none.  I can try to experiment with enabling the other tests.

> Can you help me understand more? Thanks!
>

I think your understanding is more or less on point.

Now for the PR, the purpose is to make the ELisp tests pass reliably on
Debian buildds of slower architectures.  Currently the only architecture
that can finish with a timeout of 60 seconds on any operation is amd64.
And when running manually, it looks like starting a REPL session takes a
long time, and requires a timeout of 5 minutes on i386, and as long as
15 minutes timeout on arm64.  Maybe racket is not well optimized on
other architectures yet.

Is the approach done in the PR acceptable?  And let me know if this can
be handled in a better way.

> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/greghendershott/racket-mode/pull/737#issuecomment-2600582369
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>

[1] https://tracker.debian.org/pkg/racket-mode

-- 
Regards,
Xiyue Deng

manphiz:
Xiyue Deng ***@***.***> writes:
>
> Now for the PR, the purpose is to make the ELisp tests pass reliably on
> Debian buildds of slower architectures.  Currently the only architecture
> that can finish with a timeout of 60 seconds on any operation is amd64.
> And when running manually, it looks like starting a REPL session takes a
> long time, and requires a timeout of 5 minutes on i386, and as long as
> 15 minutes timeout on arm64.  Maybe racket is not well optimized on
> other architectures yet.
>
> Is the approach done in the PR acceptable?  And let me know if this can
> be handled in a better way.
>

Actually, it probably helps to disable the timeout altogether so that on
slower architectures we just wait for the tests to finish.  Is this
possible?

-- 
Regards,
Xiyue Deng

PR #748, opened 2025-04-12
Support command line arguments to racket-program/backend

bkhl:
The documentation already had examples of configurations like

```elisp
(racket-add-back-end "/ssh:headless:~/gui-project/"
                     :racket-program "xvfb-run racket")
```

This would (accidentally?) work for remote backends, because the way the SSH
command is invoked ends up applying shell word splitting on the program.

However, when I wanted to use this locally something like this:

```elisp
(racket-add-backend "/project/"
                    :racket-program "podman
                                     run --rm --interactive
                                     --volume=/project/:/project/:z
                                     --workdir=/project/
                                     docker.io/racket/racket/8.9-full"
                                     "racket")
```

This wouldn't work because the `executable-find' call would try to find the path
to an executable with that full string as a name.

We could in theory apply word splitting on this string ourselves, but I think
it's more common for modes to have a `*-flags' variable in these cases, so I add
that here.

### DIFF
--- a/racket-custom.el
+++ b/racket-custom.el
@@ -40,12 +40,20 @@
 (defvar racket--winp (eq 'windows-nt system-type))
 
 (defcustom racket-program (if racket--winp "Racket.exe" "racket")
-  "Pathname of the Racket executable.
+  "Pathname of the Racket executable or command line to launch it.
+
+- If the value of this variable is a string, it will be interpreted as a
+  simple command without arguments.
+
+- If it is a list of strings, the first element will be taken to be the
+  executable, and the rest of the list command line arguments to pass to
+  it before any other arguments.
 
 Note that a back end configuration can override this with a
 non-nil `racket-program` property list value. See
 `racket-add-back-end'."
-  :type '(file :must-match t)
+  :type '(choice (string)
+                 (repeat string))
   :risky t)
 
 (make-obsolete-variable 'racket-command-port nil "2020-04-25")
--- a/racket-doc.el
+++ b/racket-doc.el
@@ -60,11 +60,15 @@ Centralizes how to issue doc command and handle response correctly."
 
 (defun racket--search-doc-locally (str)
   (racket--doc-assert-local-back-end)
-  (call-process racket-program
-                nil ;INFILE: none
-                0   ;DESTINATION: discard/don't wait
-                nil ;DISPLAY: none
-                "-l" "raco" "docs" str))
+  (let ((command (if (stringp racket-program)
+                     (list racket-program)
+                   racket-program)))
+    (apply #'call-process `(,(car command)
+                            nil ;INFILE: none
+                            0   ;DESTINATION: discard/don't wait
+                            nil ;DISPLAY: none
+                            ,@(cdr command)
+                            "-l" "raco" "docs" ,str))))
 
 (provide 'racket-doc)
 
--- a/racket-shell.el
+++ b/racket-shell.el
@@ -11,6 +11,7 @@
 (require 'racket-custom)
 (require 'racket-util)
 (require 'shell)
+(require 'subr-x)
 (require 'term)
 
 (defun racket-racket ()
@@ -34,11 +35,17 @@ variable `racket-shell-or-terminal-function'."
 
 (defun racket--shell-or-terminal (args)
   (racket--save-if-changed)
-  (let* ((exe (shell-quote-argument
-               (if (file-name-absolute-p racket-program)
-                   (expand-file-name racket-program) ;handle e.g. ~/
-                 racket-program)))
-         (cmd (concat exe " " args))
+  (let* ((command (if (stringp racket-program)
+                      (list racket-program)
+                    racket-program))
+         (program (car command))
+         (exe (shell-quote-argument
+               (if (file-name-absolute-p program)
+                   (expand-file-name program) ;handle e.g. ~/
+                 program)))
+         (flags (mapcar (lambda (x) (shell-quote-argument x))
+                        (cdr command)))
+         (cmd (concat exe " " (string-join flags " ") args))
          (win (selected-window)))
     (funcall racket-shell-or-terminal-function cmd)
     (select-window win)))
--- a/doc/racket-mode.texi
+++ b/doc/racket-mode.texi
@@ -3147,7 +3147,18 @@ Delete the ``compiled'' directories made by @ref{racket-mode-start-faster}.
 @node racket-program
 @subsection racket-program
 
-Pathname of the Racket executable.
+Pathname of the Racket executable or command line to launch it.
+
+@itemize
+@item
+If the value of this variable is a string, it will be interpreted as a
+simple command without arguments.
+
+@item
+If it is a list of strings, the first element will be taken to be the
+executable, and the rest of the list command line arguments to pass to
+it before any other arguments.
+@end itemize
 
 Note that a back end configuration can override this with a
 non-nil @code{racket-program} property list value. See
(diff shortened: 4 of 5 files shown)


greghendershott:
1. Thanks for the commit to fix the misleading error message! That message must have made things very confusing for you -- sorry!

2. Thanks for the commit for invoking a program with a list of args/flags. But... I need to mull this over a bit...

Originally I made this `racket-program` -- solely a path to an executable -- partly because I didn't want people to think they could/should invoke `racket` with flags _for Racket_ that might be incompatible with what the back end needs. So I deliberately avoided the idea of it being a command string, or, an executable plus a list of args. Instead it's supposed to be just something like "racket", "Racket.exe", or "some/path/to/racket".

But obviously I lost the plot with the `xvfb-run racket` example.

Anyway, that example suggests a broader use case I never considered originally -- where something "wraps" the invocation of Racket, such as xvfb-run or podman as in your motivating example.

---

My first reaction (might change):

I am leaning against adding a whole new "racket-flags" variable and back end key.

Instead, I'm leaning toward keeping `racket-program` and either:

- do the word splitting explicitly, always (not just rely on tramp auto splitting for remote back ends)

OR

- allow `racket-program` to be a string (as now) or also (new) a list of strings, i.e. the exec plus zero or more args/flags

(With either option, `racket-program` in hindsight would better be named something like `racket-command`. But I think just leave the name as-is and explain in the docs.)

What do you think?

bkhl:
I haven't done a massive survey, but I've seen packages before that just allow a string with arguments in, and some that have a separate `flags` type variable for similar uses, but none with the "string or list" approach, so I'd rather go for either the whole command in a single string, or a "flags" parameter.

Also not sure if the customize UI can even handle an option that can be one of two different types?

If you think the flags are awkward I have nothing against either approach.

If we go with the single string approach, we could use `split-string-shell-command` if we don't want to implement the parsing quoted parameters ourselves, and if you don't mind a dependency on `shell`?

Keeping the option as it but word-splitting in Emacs also solve the security implication that might otherwise be there, since we can then still wordsplit the string but "exec" it locally without passing it through a shell. (For the remote case over SSH that's not an option of course).

bkhl:
@greghendershott , I just pushed a new version that does shell splitting of the command instead, but have not tested it at all. Let me know if you approve of the general approach. If you do I'll do a bit of testing and also try to get the CI pipeline to pass.

greghendershott:
> I just pushed a new version that does shell splitting of the command instead, but have not tested it at all. Let me know if you approve of the general approach. If you do I'll do a bit of testing and also try to get the CI pipeline to pass.

Thanks!

Although I'd be OK with the value being a single string, that we split, from the CI failures it looks like `split-string-shell-command` is undefined in Emacs prior to 29.1: https://github.com/greghendershott/racket-mode/actions/runs/14427680723/job/40461511981?pr=748#step:7:66.

Because I try to support somewhat older Emacs (ergo those CI jobs), we can't depend on something that new. :disappointed: 

A quick glance at its implementation makes me feel like we shouldn't try to copypasta it.

(At least not copy it verbatim. It seems to call `shell--parse-pcomplete-arguments` which in turn calls `shell-backward-command`, i.e. move in an interactive shell buffer -- which seems like the wrong thing. Maybe I'm mistaken if it's working in your testing, but it seems wrong? Unless it's stuffing the string in a temporary shell buffer, that I overlooked.)

---

So: Letting `racket-program` be a string, or, a list of strings, feels like the least-worst approach, to me. Is that OK for you?

Although I agree it's probably less-common for _customization_ variables, Emacs generally has many examples like this -- such as "The value is a string, or a number, or a list of either, or a function that returns the preceding". :smile: The customization DSL for `:type` can handle more than you might think, via combining clauses like `or` and `repeat`. I'm 95% sure it can handle this. Also, if you wanted to leave that last detail to me, to do, I wouldn't object.

---

p.s. The CI run for Emacs 29.1 seems like straightforward byte compiler warnings about lexical vars. You can get these locally with `make compile clean` in the `racket-mode` directory.

bkhl:
Alright, I've made something like this and again it works at least with `racket-run-and-switch-to-repl` with a `.dir-locals.el` like this:

```elisp
((racket-mode
  . ((eval
      . (let ((root (project-root (project-current)))
              (user-emacs-directory (expand-file-name user-emacs-directory)))
          (racket-add-back-end root
                               :racket-program `("podman" "run" "--rm" "--interactive"
                                                 ,(concat "--volume="
                                                          user-emacs-directory
                                                          ":"
                                                          user-emacs-directory
                                                          ":z")
			                         ,(concat "--volume=" root ":" root ":z")
			                         ,(concat "--workdir=" root)
			                         "localhost/racket:latest"
                                                 "racket"))
          (racket-repl-buffer-name-shared))))))
```

(Found that in `.dir-locals.el` I need to call the function to set the buffer name to make it use the new backend first time i load `racket-mode`, but that's maybe the subject for another PR …)

I'm a bit more unsure of the use-cases of the functions in `racket-doc` and `racket-shell`, so would appreciate if you could take an extra good look on those.

Also a PS about the `racket-program` variable. Does it really make sense to define it as `file :must-match t`, when it might refer to a remote backend where the customize interface would not be able to give you a file selector?

racket-custom.el
```diff
@@ -40,12 +40,20 @@
...
+
+- If it is a list of strings, the first element will be taken to be the
+  executable, and the rest of the list command line arguments to pass to
+  it before any other arguments.
 
 Note that a back end configuration can override this with a
 non-nil `racket-program` property list value. See
 `racket-add-back-end'."
-  :type '(file :must-match t)
+  :type '(choice (file :must-match t)
```
greghendershott, line 55: You asked if `(file :must-match t)` is correct. I agree, it's _not_ correct. Just `string` is right.

racket-shell.el
```diff
@@ -34,11 +35,17 @@ variable `racket-shell-or-terminal-function'."
...
+  (let* ((command (if (stringp racket-program)
+                      (list racket-program)
+                    racket-program))
+         (program (car command))
+         (exe (shell-quote-argument
+               (if (file-name-absolute-p program)
+                   (expand-file-name program) ;handle e.g. ~/
+                 program)))
+         (flags (mapcar (lambda (x) (shell-quote-argument x))
+                        command))
```
greghendershott, line 47: Instead of `command` should this be `(cdr command)` to get just the flags?

racket-doc.el
```diff
@@ -60,11 +60,15 @@ Centralizes how to issue doc command and handle response correctly."
...
-                "-l" "raco" "docs" str))
+  (let ((command (if (stringp racket-program)
+                     (list racket-program)
+                   racket-program)))
+    (apply call-process `(,(car command)
+                          nil ;INFILE: none
+                          0   ;DESTINATION: discard/don't wait
+                          nil ;DISPLAY: none
+                          ,@(cdr command)
+                          "-l" "raco" "docs" str))))
```
greghendershott, line 71: We need to unquasiquote `str` here, i.e. add a `,` before it. That's what the byte compiler is complaining about on CI: ` racket-doc.el:61:1:Warning: Unused lexical argument ‘str’`.

greghendershott:
Thank you!

I just left 4 comments, which, AFAICT should make this good to go.

greghendershott:
@bkhl By the way, I appreciate all the work you've done so far! Probably you're very busy. Maybe you only have spare time on the weekends.

I'm happy to wait for you to finish this; it seems very, very close now. Or, if you prefer, I can do the final few steps. It's 100% your choice, either way. Just let me know, eventually, at your convenience.

greghendershott:
Merged -- thanks!!

PR #750, opened 2025-04-25
hash-lang: lex using 'more-specific-type when available

Bogdanp:
Together with https://github.com/racket/rhombus/pull/666, this change brings font locking for Rhombus closer to Racket. Happy to take a different approach if you think there's a better way to do this! As I mentioned in the other PR, I don't love calling the key "more-specific-type" so suggestions welcome.


greghendershott:
Merged; thanks!

PR #752, opened 2025-05-28
repair racket-project-root

capfredf:
The representation of a project seems to have changed.  `cdr (project nil dir))` returns `(VC <dir>)` rather than the directory wanted. `project-root` hides the details and gives us the desirable result.

### DIFF
--- a/racket-util.el
+++ b/racket-util.el
@@ -118,7 +118,9 @@ The \"project\" is determined by trying, in order:
         (and (fboundp 'vc-root-dir)
              (vc-root-dir))
         (and (fboundp 'project-current)
-             (cdr (project-current nil dir)))
+             (let ((pr (project-current nil dir)))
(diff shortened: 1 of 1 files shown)


racket-util.el
```diff
@@ -118,7 +118,9 @@ The \"project\" is determined by trying, in order:
         (and (fboundp 'vc-root-dir)
              (vc-root-dir))
         (and (fboundp 'project-current)
-             (cdr (project-current nil dir)))
+             (let ((pr (project-current nil dir)))
+               (and (fboundp 'project-root)
```
capfredf, line 122: @greghendershott I am not sure this is the right way to make CI happy.

capfredf: I think what we really want here is to load the library `project` on demand if it exists, but I have not figured how to do that properly.

